PR #14, opened 2023-01-18
Spelling 😅

RaeesBhatti:
Changes `we can have our cake and eat it to!` to `we can have our cake and eat it too!`


sam-goodwin:
LOL! Thanks :)

PR #19, opened 2023-01-21
feat: add S3 support

### LINKED ISSUE #3: S3 support

naorpeled:
Closes #3

### DIFF
--- a/test/s3.test.ts
+++ b/test/s3.test.ts
@@ -7,13 +7,13 @@ test("S3 PutObject and GetObject should work", async () => {
   const Body = "test-body";
 
   await S3.putObject({
-    Bucket: "naor-python",
+    Bucket: "itty-aws",
     Key,
     Body,
   });
 
   const response = await S3.getObject({
-    Bucket: "naor-python",
+    Bucket: "itty-aws",
     Key,
   });
 
--- a/test/s3.test.ts
+++ b/test/s3.test.ts
@@ -1,0 +1,21 @@
+import { AWS } from "../src";
+
+const S3 = new AWS.S3();
+
+test("S3 PutObject and GetObject should work", async () => {
+  const Key = "test-key";
+  const Body = "test-body";
+
+  await S3.putObject({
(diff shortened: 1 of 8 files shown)


test/s3.test.ts
```diff
@@ -0,0 +1,21 @@
...
+    Key,
+    Body,
+  });
+
+  const response = await S3.getObject({
+    Bucket: "itty-s3-bucket",
+    Key,
+  });
+
+  expect(response?.Body?.toString()).toEqual(Body);
```
sam-goodwin, line 20: I wonder how well this works with Streaming IO? We will have to test it with a big file - it's important to be able to stream io in and out rather than rely on the entire object always fitting in memory.

sam-goodwin (commented):
This looks great. Nice find on the xml library, looks like a minimal impact on bundle size and is browser compatible!

Main concern is around streaming IO for large files. Both inputs and outputs - we can't rely on everything being a string.

naorpeled:
> This looks great. Nice find on the xml library, looks like a minimal impact on bundle size and is browser compatible!
> 
> Main concern is around streaming IO for large files. Both inputs and outputs - we can't rely on everything being a string.

Good point, will play around with it and see what adjustments must be done to make it work 😎

sam-goodwin:
Any updates here?

naorpeled:
> Any updates here? 

Will finish this on Friday probably, I'm duty at work this week so couldn't continue working on this  😢

sam-goodwin:
Ah right, makes sense lol. Sorry! Thanks for contributing while also being busy :)

naorpeled:
> Ah right, makes sense lol. Sorry! Thanks for contributing while also being busy :)

All good, I'm sorry it's taking longer anticipated

sam-goodwin:
Pushed some changes - S3 is a weird AF API. Lots of heuristics around headers was required. I think we can just prioritize making it work for high important APIs, like getObject, putObject, listObjectsV2. etc.

PR #24, opened 2023-03-31
chore: add "main" to package.json to fix compatibility with some frameworks

yehudacohen:
…eworks


sam-goodwin (approved):
thanks for catching this!

PR #34, opened 2024-01-31
feat: add SQS support

### LINKED ISSUE #1: SQS support

### LINKED ISSUE #2: SNS support

skyrpex:
- Resolves #1
- Fix pnpm workspace
- Fix `bench:synth`
- Use modern tooling (node 20, tsup, tsx, vitest, ...)
- Use LocalStack to run tests


sam-goodwin:
Wow, thanks for the large change. This is awesome!

skyrpex:
> Wow, thanks for the large change. This is awesome!

No problem! Let me know if you want me to change something, or feel free to do it yourself.

aws-models/accessanalyzer.json
```diff
@@ -1,6 +1,24 @@
 {
   "smithy": "2.0",
   "shapes": {
+    "com.amazonaws.accessanalyzer#Access": {
```
sam-goodwin, line 4: I notice we are downloading the updated models but I can't remember if I was ever using them. I think I am still using the AWS SDK v2 to get the types (a lazy approach) but wanted to generate them from the models.

skyrpex: Oh, I have no idea either. Will try to have a look soon.

skyrpex:
I'll check if SQS and SNS clients work...

skyrpex:
Hey so it still doesn't work for SNS and SQS. What does it take for it to work, is it on AWS' side? Sorry about the hassle 😢

sam-goodwin:
We will have to check that it is sending the right protocol. Does the JSON api as defined in the types match what the api expects?

No bother! Sorry I haven't been able to help, have been busy!

skyrpex:
Spent some hours last night diving into this. Got good news:

- SQS works without additional work 💪🏻 
- SNS still uses the XML protocol thingy, but can be adapted
- Prepared the tests to work with LocalStack
- I think LocalStack has a bug with SNS.publishBatch

I was thinking about repurposing this PR so it runs tests in CI with LocalStack, including SQS. Then, add support for SNS with a custom SNS handler.

sam-goodwin:
Awesome! That's interesting about SNS, what needs to be done to adapt?

Should we re open the PR?

skyrpex:
SNS seems to work the same way as S3 does:

- Must use POST
- Must pass `Action` as a query param: `url.searchParams.set("Action", resolveAction(methodName));`, and the rest of the inputs too
  - A notable exception is https://docs.aws.amazon.com/sns/latest/api/API_PublishBatch.html#API_PublishBatch_RequestParameters which uses a weird array format `PublishBatchRequestEntries.member.N`
- Returns XML but we should be able to parse it

sam-goodwin:
Ah, got it. So only SQS received the JSON API upgrade. Here's hoping 🤞 they do this across the board.

sam-goodwin:
SNS api is probably a lot simpler than S3? S3 had complicated surface area and streaming protocol on top of that.

skyrpex:
> SNS api is probably a lot simpler than S3? S3 had complicated surface area and streaming protocol on top of that.

I bet! I'll have a look on SNS soon.

sam-goodwin:
Wanna merge SQS? We can release incrementally without much worry.

sam-goodwin:
Don't think I've ever seen a million line PR. Is that because we are checking in AWS models? Maybe we should not check that in?

skyrpex:
> Don't think I've ever seen a million line PR. Is that because we are checking in AWS models? Maybe we should not check that in?

Yup, done 👍🏻 

> Wanna merge SQS?

Sure!

scripts/download-smithy-models.sh
```diff
@@ -1,5 +1,6 @@
-git clone git@github.com:aws/aws-sdk-js-v3.git --depth=1
+git clone https://github.com/aws/aws-sdk-js-v3.git --depth=1
```
sam-goodwin, line 1: Why https protocol? I don't have strong opinions but was this done for a specific reason?

sam-goodwin (commented):
Hard to identify changes because of formatting updates. But I'm also glad you updated the repo's config so thoroughly lol.

scripts/download-smithy-models.sh
skyrpex: I'm not sure why but the git protocol didn't work for me. I will revert that too since I don't really have to use it.

skyrpex:
> Hard to identify changes because of formatting updates. But I'm also glad you updated the repo's config so thoroughly lol.

Important changes to `src/index.ts` are:

- Endpoint resolution. It assumed custom endpoints didn't include the protocol and couldn't use `http:/` (eg, localstack)
- Region resolution. It now supports `options.region` and doesn't crash if env.AWS_REGION isn't defined

The rest is just formatting changes.

README.md
```diff
@@ -33,8 +33,8 @@ This project aims to eliminate the following issues with the official AWS SDK:
 
 The entire AWS SDK (including all Services and APIs) fits in to a
 
-- Minified bundle size of: `49 KB`.
-- Un-minified bundle size of: `95 KB`.
+- Minified bundle size of: `22 KB`.
+- Un-minified bundle size of: `32 KB`.
```
sam-goodwin, line 37: Awesome!

sam-goodwin (approved):

PR #37, opened 2025-07-29
use dynamic Content-Type headers based on AWS service protocol

kmitchener:
Makes Content-Type headers dynamic based on each AWS service's protocol specification.

**Problem**: All services were using hardcoded `application/x-amz-json-1.0`, but CloudTrail requires `application/x-amz-json-1.1` and other services use different protocols entirely.

**Changes**:
- Enhanced protocol detection to distinguish between `awsJson1_0`, `awsJson1_1`, `restJson1`, `awsQuery`, `ec2Query`, and `restXml`
- Updated client to set appropriate Content-Type based on service protocol:
  - `awsJson1_0` → `application/x-amz-json-1.0` (DynamoDB, SQS, etc.)
  - `awsJson1_1` → `application/x-amz-json-1.1` (CloudTrail, Rekognition, etc.)
  - `restJson1` → `application/json` (Lambda, S3, etc.)
  - Plus support for form-encoded and XML protocols

**Impact**: Fixes compatibility issues with services that require specific Content-Type headers.


sam-goodwin:
This is brilliant, thank you for diving in and helping debug.

PR #39, opened 2025-07-30
improve dx - remove type errors for itty-aws users

kmitchener:
This change removes the red squigglies, adds completion for valid AWS services and their methods

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,20 +1,11 @@
 import * as Data from "effect/Data";
 import { createServiceProxy } from "./client.ts";
 
-import type * as Services from "./aws.ts";
-import type { AWSClientConfig } from "./client.ts";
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -1,11 +1,20 @@
...
 const services: Record<string, Record<string, any>> = {};
 
-export type AWS = typeof AWS_;
+// Create constructor types for all AWS services that return the actual service instance
+type ServiceConstructor<T> = new (config?: AWSClientConfig) => T;
+
+// Transform exported services into constructors that return the proper service types
+type AWSServices = {
+  [K in keyof typeof Services]: ServiceConstructor<InstanceType<typeof Services[K]>>
+};
```
sam-goodwin, line 15: Can we instead fix the generated types? I thought this was working.

PR #46, opened 2025-08-06
feat: fully working ec2query protocol (bidirectional)

kmitchener:
This is a much improved ec2query protocol handler. This is _not_ optimized yet to reduce the amount of metadata we're holding in memory, but it will be much easier to optimize than the previous version. 

In addition to a better implementation, this one is actually fully working for input / output translation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "test:run": "vitest run",
     "test:watch": "vitest watch",
     "test:ui": "vitest --ui",
-    "test:protocol": "bun run build && vitest run ./test/ec2-protocol.test.ts"
+    "test:protocol": "bun run build && vitest run ./test/protocols/*.test.ts"
   },
   "dependencies": {
     "aws4fetch": "^1.0.20"
--- a/test/protocols/ec2-protocol.test.ts
+++ b/test/protocols/ec2-protocol.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "@effect/vitest";
-import { Ec2QueryHandler } from "../dist/protocols/ec2-query.js";
+import { Ec2QueryHandler } from "../../dist/protocols/ec2-query.js";
 
 const handler = new Ec2QueryHandler();
 


src/protocols/ec2-query.ts
```diff
@@ -1,21 +1,238 @@
...
+    .then(({ ec2ModelMeta: meta }) => {
+      ec2ModelMeta = meta;
+    })
+    .catch(() => {
+      // Metadata not available, will use fallback
+      ec2ModelMeta = null;
+    });
+} catch {
+  // Module not found or error during import
+}
```
sam-goodwin, line 19: Why do we want to swallow this error?

src/protocols/ec2-query.ts
```diff
@@ -1,21 +1,238 @@
...
+          toParams(ec2ModelMeta.shapes, op.input, input, "", enhancedParams);
+        }
+
+        const usp = new URLSearchParams();
+        for (const [k, v] of Object.entries(enhancedParams)) usp.append(k, v);
+        return usp.toString();
+      }
+    } else {
+      //FIXME: should throw error
+    }
```
sam-goodwin, line 235: What's the story here?

src/protocols/ec2-query.ts
kmitchener: yeah, we don't want to swallow that. it previously had some junk as "fallback", and fallback isn't possible. I needed to research more about dynamic imports and how they worked ... anyway, updated it now to throw a hard error if it can't load the metadata.

src/protocols/ec2-query.ts
kmitchener: resolved with change to error handling above

kmitchener:
protocol tests added (from https://github.com/smithy-lang/smithy/tree/main/smithy-aws-protocol-tests/model/ec2Query). the attempt to dynamically import the ec2-metadata is removed. we will address that issue when we split the services in https://github.com/sam-goodwin/itty-aws/issues/41

PR #52, opened 2025-08-11
fix: ec2Query input transformation bug

kmitchener:
Fixes a bug in EC2 Query protocol input transformation where field names and list flattening weren't handled correctly.

  Changes:
  - Fixed field name capitalization for EC2 Query parameters
  - Simplified list flattening to use correct index format (prefix.1, prefix.2)
  - Removed lazy metadata loading in favor of direct import
  - Added comprehensive EC2 smoke tests covering VPC lifecycle operations

  Test Coverage:
  - VPC creation, state monitoring, and deletion
  - Error handling for non-existent resources
  - Invalid parameter validation

### DIFF
--- a/src/protocols/ec2-query.ts
+++ b/src/protocols/ec2-query.ts
@@ -17,6 +17,10 @@ function safeParseXml(xmlText: string): any {
   }
 }
 
+function capitalizeFirst(str: string): string {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
+
 function toParams(
   shapes: Record<string, any>,
   shapeId: string,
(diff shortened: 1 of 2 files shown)


src/protocols/ec2-query.ts
```diff
@@ -28,6 +17,10 @@ function safeParseXml(xmlText: string): any {
   }
 }
 
+function capitalizeFirst(str: string): string {
+  return str.charAt(0).toUpperCase() + str.slice(1);
+}
```
sam-goodwin, line 22: might be worth collecting `src/utils`

`src/utils/captialize.ts`

sam-goodwin (approved):

PR #74, opened 2025-09-27
generate service-specific errors union type

stackops-ngin:
simplifying addition of tagged errors by service


stackops-ngin:
`git submodule update --remote aws-models` and `bun run generate` done and committed - hope this helps!

PR #78, opened 2025-11-11
feat: auto retry certain errors

Mkassabov:


src/client.ts
```diff
@@ -247,7 +257,21 @@ export function createServiceProxy<T>(
...
+              }),
+            ),
+            Effect.retry({
+              schedule: Schedule.intersect(
+                Schedule.exponential("100 millis"),
+                Schedule.recurs(5),
+              ),
+              while: (e) => retryableErrorTags.includes(e._tag),
+            }),
+          );
```
sam-goodwin: Is there a way we can make this configurable or injectable as a layer with a sensible default.

src/client.ts
```diff
@@ -6,6 +6,16 @@ import { Credentials, fromStaticCredentials } from "./credentials.ts";
...
+import * as Schedule from "effect/Schedule";
+
+export const retryableErrorTags = [
+  "InternalFailure",
+  "RequestExpired",
+  "ServiceException",
+  "ServiceUnavailable",
+  "ThrottlingException",
+  "TooManyRequestsException",
+];
```
sam-goodwin: Is this all of the errors? Can we extract them. Maybe the smithy models has information we cam use, e.g. all 5xx errors should be retried. Maybe there's other metadata.

PR #185, opened 2026-04-10
Migrate distilled to Effect 4 context services

joepjoosten:
## Summary
- update the repo to the new Effect 4 beta versions in package.json and bun.lock
- replace removed ServiceMap usage with Context.Service across SDK service tags
- replace Effect.services() captures with Effect.context() and Effect.provideContext(...)

## Verification
- bun run typecheck
- bun run typecheck (packages/aws)
- bun run typecheck (packages/gcp)
- bun run typecheck (packages/kubernetes)


Mkassabov:
seems we have some problems with our test suite rn. going to fix those, then run your changes against tests. TBH I don't see why this shouldn't merge, and we appreciate the work!

We also might delay a little bit as we currently pin all effect version across distilled and alchemy-effect so we need to make sure alchemy-effect gets upgraded to effect4 beta 46 as well.

Mkassabov:
merged thank you for your contribution; release will follow shortly

PR #192, opened 2026-04-13
John/cf queue fix

john-royal:


john-royal:
Broken up into separate commits

PR #199, opened 2026-04-29
fix: Add missing extensions on module imports

lucasthevenet:
This PR adds .ts file extensions to local imports on the Planetscale package that were currently missing. 

Under certain ESM configs having no file extension on these files breaks module resolution as seen here:

Screenshot 2026-04-28 at 5 03 51 PM

The missing file extensions might be an issue in other packages as well, but I didn't want to touch more than necessary.


sam-goodwin:
This looks good to me. @Mkassabov can you make sure this is consistent across all of distilled services.

sam-goodwin:
Oh, looks like the generator script was not modified, only the generated source. We should fix the generator script @lucasthevenet

lucasthevenet:
Sure! Jus fixed the generator script and all the providers that had been generated incorrectly

PR #207, opened 2026-05-01
fix(core): auto-retry HTTP 429 with capped exponential backoff

Mkassabov:
Several SDKs (notably posthog) get rate-limited by their upstream APIs when the test suite runs in CI under contention. The shared client had no retry behaviour at all — \`matchError\` immediately turned 429 into a typed \`TooManyRequests\`, which then cascaded into hundreds of speculative-tag-assertion failures (\`expected 'TooManyRequests' to be 'NotFound'\`).

Wrap the per-operation Effect with an automatic retry on a sentinel that's raised internally for status 429. Schedule is exponential (1s base, factor 2) capped at 8 attempts via \`Schedule.both(...)\`. After retries are exhausted the original error body is fed to the SDK's \`matchError\` so callers still see the typed 429 outcome, just with the benefit of having waited the rate limit out first.

Doesn't currently honour Retry-After headers (would need a per-attempt \`Ref\` to thread the value into the schedule); the exponential ramp covers the rate limits we actually encounter. Affects every SDK that uses \`makeAPI\` from core.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -25,8 +25,10 @@
  * const result = yield* fn({ organization: "my-org" });
  * ```
  */
+import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import { pipeArguments } from "effect/Pipeable";
+import * as Schedule from "effect/Schedule";
 import * as Schema from "effect/Schema";
 import * as AST from "effect/SchemaAST";
 import * as Stream from "effect/Stream";
@@ -41,6 +43,7 @@ import {
   type PaginatedTrait,
   type PaginationStrategy,
 } from "./pagination.ts";
+import * as Category from "./category.ts";
 import * as Traits from "./traits.ts";
 import { getPath } from "./traits.ts";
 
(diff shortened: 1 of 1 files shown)


Mkassabov:
shouldn't this come from sdks themselves. we shouldhave a more generic 429 that we can detect across all sdks (maybe have that error in code) then retry on that error rather than having a sentinel

Mkassabov:
> shouldn't this come from sdks themselves. we shouldhave a more generic 429 that we can detect across all sdks (maybe have that error in code) then retry on that error rather than having a sentinel

You're right — the infra is already there, I was just bypassing it. Pushed a refactor that:

- Drops the sentinel entirely.
- Lets the SDK's `matchError` do its normal job (every SDK already maps 429 → `TooManyRequests` via `HTTP_STATUS_MAP`, and that class is already decorated with `Category.withThrottlingError` + `Category.withRetryable({ throttling: true })`).
- Wraps the per-operation Effect with `Effect.retry({ while: Category.isThrottling, schedule })`, so any error tagged with the throttling category retries — no special-casing of HTTP status anywhere in the client.

That means SDK-specific subclasses (e.g. a hypothetical `CloudflareRateLimited extends TooManyRequests` decorated with the same category trait) get retried automatically without any client-side change.

Diff is much smaller now:

```ts
const fn = (input) =>
  innerFn(input).pipe(
    Effect.retry({
      schedule: throttlingRetrySchedule,
      while: (e) => Category.isThrottling(e),
    }),
  );
```

— claude

PR #213, opened 2026-05-01
fix(core): unwrap Redacted values during request input encoding

Mkassabov:
Sensitive response fields surface as `Redacted<string>` so they don't land in logs by accident. Those same values are sometimes the input to a follow-up request — WorkOS's password reset flow returns a redacted `password_reset_token` whose type is `Redacted<string>`, then the caller hands it straight into `UserlandUsersControllerResetPassword({ token })` where `token` is declared as plain `Schema.String`.

Pre-fix the encoder saw the `Redacted`, ran its toString into the schema parser, and bailed with `Expected string, got <redacted>` — the test never even reached the wire.

Walk the input one pass before `Schema.encodeSync` and convert every `Redacted<T>` to its underlying `T` (recursively, into arrays and nested objects). Sensitive-marked fields are unaffected: by the time their schema's encode transform runs the value is already the plain underlying type, which is what the wire wants anyway.

```ts
// Before:
ResetPassword({ token: tokenResult.token.password_reset_token })
//                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                     Redacted<string> — encoder rejected this with
//                     `Expected string, got <redacted>`.

// After: same call site works, plain underlying string is what hits the wire.
```

Spotted while triaging the workos suite (#212 follow-up); noted in #211 as one of the parse-failures the strip-required patch couldn't fix.

### DIFF
--- a/packages/core/src/traits.ts
+++ b/packages/core/src/traits.ts
@@ -19,6 +19,7 @@
  * );
  * ```
  */
+import * as Redacted from "effect/Redacted";
 import * as Schema from "effect/Schema";
 import * as AST from "effect/SchemaAST";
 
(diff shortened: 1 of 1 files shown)


Mkassabov:
so if I understand this correctly this essentially allows us to pass `Recated<string>` as an input that expects to be `<string>` so that value remains redacted all the way through

Mkassabov:
> so if I understand this correctly this essentially allows us to pass `Recated<string>` as an input that expects to be `<string>` so that value remains redacted all the way through

Yes, exactly. The schema declares the field as `Schema.String` — at the type level callers still pass a plain `string`. The fix is purely runtime: before this PR, if a caller happened to pass a `Redacted<string>` (e.g. forwarding a credential), `Schema.encodeSync` would serialise it to `"<redacted>"` and the request would silently send that literal string. After this PR, `unwrapRedactedDeep` walks the input first and unwraps any `Redacted` values, so the real string flows into the encoder. The redaction guarantees still hold in logs/errors because the request body assembly is what consumes it, not anything that prints it.

— claude

PR #215, opened 2026-05-01
test(supabase): gate organization-creation tests behind SUPABASE_RUN_ORGANIZATION_CREATION_TESTS

Mkassabov:
## Summary
- Supabase's management API has no `DELETE /v1/organizations/{slug}` endpoint — every organization a test creates is permanent in the user's account, and the nuke script can only enumerate orgs, not delete them.
- Adds `packages/supabase/tests/organizations.test.ts` with `v1ListAllOrganizations` coverage that runs always, plus a `v1CreateAnOrganization` happy-path test gated behind a new env var.
- The gated test uses `test.skipIf(!process.env.SUPABASE_RUN_ORGANIZATION_CREATION_TESTS)` so it's skipped by default; opt in with `SUPABASE_RUN_ORGANIZATION_CREATION_TESTS=1 bun test`.

## Test plan
- [ ] `bun test` in `packages/supabase` — gated test should be skipped
- [ ] `SUPABASE_RUN_ORGANIZATION_CREATION_TESTS=1 bun test` — gated test runs (will create a permanent org)

— claude

### DIFF
--- a/packages/supabase/tests/organizations.test.ts
+++ b/packages/supabase/tests/organizations.test.ts
@@ -1,0 +1,69 @@
+import { Effect, Layer } from "effect";
+import * as Redacted from "effect/Redacted";
+import * as FetchHttpClient from "effect/unstable/http/FetchHttpClient";
+import { describe, expect, it } from "vitest";
+import { runEffect, testRunId } from "./setup";
+import { v1CreateAnOrganization } from "../src/operations/v1CreateAnOrganization";
+import { v1ListAllOrganizations } from "../src/operations/v1ListAllOrganizations";
+import { Credentials, DEFAULT_API_BASE_URL } from "../src/credentials";
+
+// Supabase's management API has no DELETE /v1/organizations/{slug} endpoint —
+// every organization created by a test run is permanent. Gate org-creation
+// tests behind an env var so CI and ordinary local runs don't accumulate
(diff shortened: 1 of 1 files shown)


Mkassabov:
why are we adding tests here? shouldn't we be looking at the existing tests that resulted in there being too many supabase organizations in an account that it caused our test suite to fail and preventing those from running

Mkassabov:
> why are we adding tests here? shouldn't we be looking at the existing tests that resulted in there being too many supabase organizations in an account that it caused our test suite to fail and preventing those from running

Investigated this and there are no existing tests in the suite that create organizations. I grepped the entire repo for `v1CreateAnOrganization` / `POST /v1/organizations` usage — only the generated operation file itself and this PR. The supabase test setup (`tests/setup.ts`) uses `v1ListAllOrganizations` to pick the first existing org and then creates *projects* under it; it never creates orgs.

So the orgs accumulating in your account aren't from this test suite — likely manual experimentation, an earlier branch, or a different tool. Two options:

1. Close this PR — there's nothing to gate.
2. Keep it as preemptive infrastructure so any future org-creation test is gated by default. The `v1ListAllOrganizations` coverage I added is also genuine new coverage we didn't have.

Happy to do either — let me know.

— claude

PR #220, opened 2026-05-01
test(supabase): disable file parallelism

Mkassabov:
Supabase's free tier caps each org at 2 active projects. When vitest runs test files in parallel, concurrent `v1CreateAProject` calls can exceed that ceiling before the previous test's deletion has propagated — surfacing as `organization members have reached their maximum limits` and failing the `v1CreateAProject` / `v1PauseAProject` / `v1RestoreAProject` happy paths in [run 25237016325](https://github.com/alchemy-run/distilled/actions/runs/25237016325).

Set `test.fileParallelism: false` so files run sequentially. Tests inside a single file already run sequentially by default; this just stops `organizations.test.ts` and `projects.test.ts` from racing each other.

```diff
+import { defineConfig } from "vitest/config";
+
 config({ path: resolve(__dirname, "../../.env") });

-export default {};
+export default defineConfig({
+  test: {
+    fileParallelism: false,
+  },
+});
```

— claude

### DIFF
--- a/packages/supabase/vitest.config.ts
+++ b/packages/supabase/vitest.config.ts
@@ -1,6 +1,16 @@
 import { config } from "dotenv";
 import { resolve } from "path";
+import { defineConfig } from "vitest/config";
 
 config({ path: resolve(__dirname, "../../.env") });
 
-export default {};
+// Supabase free-tier accounts have a 2-active-project ceiling. Running test
+// files in parallel means concurrent v1CreateAProject calls can blow that
+// ceiling before the previous test's deletion has propagated, surfacing as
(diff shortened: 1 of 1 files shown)


Mkassabov:
Superseded by #221 — file-level parallelism wasn't the actual root cause. The flake is Supabase's leaky deletion: v1DeleteAProject returns immediately but the project lingers in REMOVED state for ~30s while still counting toward the free-tier ceiling, so a previous CI run can leave the next one unable to create projects regardless of in-run parallelism. #221 tags the error, sweeps + retries from inside tests, and runs nuke as a CI pre-step. — claude

PR #221, opened 2026-05-02
fix(supabase): tag FreeProjectLimitReached and recover in tests + CI nuke

Mkassabov:
Supabase's management API has two missing primitives that make CI runs flaky:

- No `DELETE /v1/organizations/{slug}`, so any org a test creates is permanent (already gated by `SUPABASE_RUN_ORGANIZATION_CREATION_TESTS`).
- `v1DeleteAProject` returns immediately but the project lingers ~30s in REMOVED state while still counting toward the free-tier 2-active-project ceiling, so a previous CI run leaves the next one unable to create projects.

### What this PR does

- Adds a typed `FreeProjectLimitReached` error in `@distilled.cloud/supabase`. There's no error code on the wire — the only signal is a literal `"active free projects"` substring of the message — so `matchError` detects it before the generic status-code map.
- Adds a `retryOnFreeProjectLimit` test helper. On the limit error it sweeps every `distilled-supabase-*` project from the account, sleeps 20s for deletes to propagate, and retries up to 3 times. Bounded so a real, unrelated quota issue still surfaces.
- Wraps every project-creating happy-path test in `projects.test.ts`.
- CI: runs `bun run nuke` before `bun run test` for supabase so the suite always starts from a clean slate. Tolerates nuke failures because the in-test recovery still kicks in. Includes a comment explaining why this is necessary.

```diff
+      - run: bun run nuke || true
+        working-directory: packages/supabase
+        env:
+          SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
       - run: bun run test
         working-directory: packages/supabase
```

Supersedes #220 — file-level parallelism wasn't the root cause, the leaky deletion semantics were.

— claude

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -284,6 +284,23 @@ jobs:
         working-directory: packages/core
       - run: bun run check
         working-directory: packages/supabase
+      # Supabase's management API is missing primitives we'd normally use
+      # to avoid leaking test resources between CI runs:
+      #   - no DELETE /v1/organizations/{slug}, so any org a test creates
+      #     is permanent (gated separately by
+      #     SUPABASE_RUN_ORGANIZATION_CREATION_TESTS).
+      #   - v1DeleteAProject returns immediately but the project lingers
+      #     for ~30s in REMOVED state while still counting toward the
+      #     free-tier 2-active-project ceiling, so a previous CI run that
+      #     was killed mid-test or finished within the propagation window
+      #     leaves the next run unable to create new projects.
+      # Nuke before each run so we start from a clean slate. Failures are
+      # tolerated — the in-test FreeProjectLimitReached recovery path
+      # sweeps again if needed.
+      - run: bun run nuke || true
+        working-directory: packages/supabase
+        env:
+          SUPABASE_ACCESS_TOKEN: ${{ secrets.SUPABASE_ACCESS_TOKEN }}
       - run: bun run test
         working-directory: packages/supabase
         env:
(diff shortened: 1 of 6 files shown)


Mkassabov:
can we just have this retry in a little bit
```
  × Projects > v1RestoreAProject > happy path - restores a paused project 79127ms
     → not paused yet
```

like the test should see a not paused yet (also lets make sure we have the correct tag for that) after some timeout; and if we hit that error 3 times in a error with larger timeouts only then should we fail

Mkassabov:
> can we just have this retry in a little bit ... if we hit that error 3 times in a error with larger timeouts only then should we fail

Done in cf71dc83. The wait-for-INACTIVE step now runs three times with escalating budgets (60s → 120s → 180s); only if all three exhaust without seeing \`status === "INACTIVE"\` does the test fail with "not paused yet". Worst case ~6 min, typical case finishes in the first 60s window. Bumped the test timeout to 600s to fit the escalation.

On the tag — the "not paused yet" string is just an in-test sentinel for the retry-while predicate (\`Effect.fail("not paused yet" as const)\`), not an SDK-level tagged error. The Supabase API doesn't surface a typed "still pausing" error class for us to discriminate on; the only signal is polling \`v1GetProject\` and looking at \`status\`. Let me know if you want me to lift this into a real \`Schema.TaggedError\` (e.g. \`ProjectNotPausedYet\`) so it can be reused outside this one test, or leave the local sentinel.

— claude

PR #225, opened 2026-05-02
refactor(core,sdks): unify Retry into a single shared Context.Service

sam-goodwin:
## Summary

Per-SDK `Retry` Context.Service tags were never actually consumed: each `packages/<sdk>/src/retry.ts` declared its own `Retry` class but the shared `makeAPI` in `@distilled.cloud/core/client` had no way to look up N different tags and just ran a hard-coded throttling-only schedule. So `Cloudflare.Retry.transient(eff)`, `Turso.Retry.policy(...)`, etc. all compiled and provided a layer — but nothing read it.

This PR keeps the per-SDK tag (so policies don't bleed across SDKs — `Cloudflare.Retry.transient` only retries Cloudflare calls, not turso underneath) and just has each SDK's client wire its own tag into `makeAPI` automatically. End users just write `myEffect.pipe(Cloudflare.Retry.transient)` — no per-SDK config to remember.

This mirrors the AWS pattern in `packages/aws/src/client/api.ts` (`outerFn` reads `Retry` with `Retry.makeDefault` as fallback).

### Changes

- **`packages/core/src/client.ts`** — `ClientConfig.retry` is now a required `Context.Key<any, RetryPolicy>`. `makeAPI` always reads the configured tag with `Effect.serviceOption` and falls back to `Retry.makeDefault` (transient/throttling/server with capped exponential backoff + jitter, 5 attempts). Removed the legacy hard-coded throttling-only branch.
- **`packages/core/src/retry.ts`** — kept `makeRetryService` as the primary API for SDKs, removed the short-lived shared `Retry` tag from the previous commit.
- **18 per-SDK `retry.ts` files** (cloudflare, turso, planetscale, neon, stripe, kubernetes, axiom, posthog, gcp, coinbase, supabase, expo-eas, fly-io, mongodb-atlas, typesense, workos, prisma-postgres, azure) — each declares its own `Retry` tag plus matching `policy`, `none`, `throttling`, `transient` helpers (added `throttling`/`transient` everywhere for parity — previously only cloudflare had them).
- **18 per-SDK `client.ts` / `client/api.ts` files** — each imports its local `Retry` and passes it as `retry: Retry as any` into the `makeAPI` config.

AWS continues to use its own client and `Retry` tag, unchanged.

### Caller-facing result

```ts
import * as Cloudflare from "@distilled.cloud/cloudflare";
import * as Turso from "@distilled.cloud/turso";

myEffect.pipe(Cloudflare.Retry.transient);
myEffect.pipe(Turso.Retry.policy({ while: isMyError, schedule: mySchedule }));
```

Every Cloudflare/Turso/etc. API call below transparently retries transient/throttling/server errors — no per-resource `Effect.retry` blocks, and policies stay scoped to each SDK.

## Test plan

- [x] tsgo typecheck clean for core, aws, and 17 of 18 SDKs (cloudflare, turso, planetscale, neon, stripe, kubernetes, axiom, posthog, coinbase, supabase, fly-io, mongodb-atlas, typesense, workos, prisma-postgres, azure, expo-eas). gcp has a pre-existing unrelated readonly-array error verified against main.
- [x] Cloudflare unit tests pass (client-api, credentials, spec-emission).
- [ ] CI: live test suites that already call `Retry.transient` in their test harnesses now exercise the real wiring instead of being no-ops.

### DIFF
--- a/packages/axiom/src/client.ts
+++ b/packages/axiom/src/client.ts
@@ -8,6 +8,7 @@ import * as Effect from "effect/Effect";
 import * as Redacted from "effect/Redacted";
 import * as Schema from "effect/Schema";
 import { makeAPI } from "@distilled.cloud/core/client";
+import { Retry } from "./retry.ts";
 import {
   HTTP_STATUS_MAP,
   UnknownAxiomError,
@@ -108,5 +109,6 @@ export const API = makeAPI<Credentials>({
   },
   matchError,
   ParseError: AxiomParseError as any,
+  retry: Retry as any,
   transformResponse: stripNulls,
 });
--- a/packages/axiom/src/retry.ts
+++ b/packages/axiom/src/retry.ts
@@ -1,22 +1,53 @@
 /**
- * axiom retry configuration.
+ * Axiom retry configuration.
  *
- * Re-exports the shared `Retry` Context.Service from `@distilled.cloud/core`
- * so a blanket retry policy installed at the layer level covers every
- * axiom API call below it.
+ * Defines the per-SDK `Retry` Context.Service tag that
+ * `packages/axiom/src/client.ts` wires into `makeAPI`. Callers can
+ * install a blanket retry policy at the layer level and have every
+ * Axiom API call below it pick it up:
+ *
+ * @example
+ * ```ts
+ * import * as Axiom from "@distilled.cloud/axiom";
+ *
+ * myEffect.pipe(Axiom.Retry.transient);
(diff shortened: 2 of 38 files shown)


Mkassabov:
Make sure create-sdk/create-sdk-full are updated so this is automatic for future sdks (just tell Claude it can update the scripts).

The tests failing rn are fine, those are broken on main. Just make sure it doesn't get worse.

Most important one is supabase since its rate limit stuff is super noticable if you break it. If that test is fine the rest should be.

Mkassabov:
retries are entirely reworked in #390

PR #228, opened 2026-05-04
feat(core): parse Retry-After / Ratelimit headers and honor them in retry policy

sam-goodwin:
## Summary

Standardizes how SDKs surface server-provided retry hints. Stacks on top of #225.

- **Core** defines an opaque `retryAfter: Duration.Duration` field on every retryable error (`TooManyRequests`, `Locked`, `InternalServerError`, `BadGateway`, `ServiceUnavailable`, `GatewayTimeout`).
- **Core** ships parsing primitives: `parseRetryAfter` (RFC 7231 §7.1.3 — `delay-seconds` or HTTP-date), `parseRatelimit` (IETF httpapi-ratelimit-headers — `r=0, t=30`), and `parseServerRetryHint` (try one then the other).
- **Each SDK's `matchError`** now accepts an optional `headers` arg and is responsible for deriving a `Duration` from whatever the service actually returns. Cloudflare, GCP, Kubernetes, plus all 15 scaffolded SDKs (turso, planetscale, neon, stripe, axiom, posthog, supabase, coinbase, fly-io, mongodb-atlas, typesense, workos, prisma-postgres, azure, expo-eas) populate `retryAfter` via `parseServerRetryHint(headers)` for the standard cases. The `scripts/create-sdk.ts` template emits the same wiring so newly scaffolded SDKs honor server hints by default.
- **Default retry policy** (`makeDefault`) and the new `throttlingFactory` / `transientFactory` honor `error.retryAfter` with precedence over the exponential backoff, capped at 60s so a misbehaving server can't park us forever.
- **AWS untouched** — uses its own retry stack.
- **Pipeline awareness**: prompts in `scripts/error-discovery.ts` and `scripts/generate-tests.ts` teach sub-agents to preserve the `headers` parameter on `matchError` and pass `retryAfter` into retryable error constructors when patching, so the convention sticks across regeneration.

### Override seam

If a service uses bespoke headers or body fields for cooldown, edit the SDK's `matchError`:

```ts
const retryAfter =
  parseServerRetryHint(headers) ??
  parseFooBespokeHeader(headers["x-foo-cooldown"]) ??
  parseFooBodyField(errorBody);
return Effect.fail(new TooManyRequests({ message, retryAfter }));
```

## Test plan

- [x] `bun run test` in `packages/core` — 29 new unit tests pass (header parsing edge cases + policy honoring `retryAfter`)
- [x] `bunx tsgo --noEmit` clean for cloudflare, kubernetes, turso, planetscale, neon, stripe, axiom, posthog, supabase, coinbase, fly-io, mongodb-atlas, typesense, workos, prisma-postgres, azure, expo-eas
- [x] `bunx oxlint src` clean across all 18 SDKs
- [x] GCP shows pre-existing readonly-array errors only (also present on `main`, out of scope)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/core/src/retry.ts
+++ b/packages/core/src/retry.ts
@@ -2,24 +2,14 @@
  * Retry Policy System
  *
  * Provides configurable retry policies for API operations.
- * Each SDK creates its own Retry service tag but uses these shared utilities
- * for building retry schedules and policies.
- *
- * @example
- * ```ts
- * import * as Retry from "@distilled.cloud/core/retry";
- *
- * // Use the default retry policy
- * myEffect.pipe(Retry.policy(myRetryService, Retry.makeDefault()))
- *
- * // Disable retries
- * myEffect.pipe(Retry.none(myRetryService))
- * ```
+ * Each SDK creates its own `Retry` Context.Service tag (via
+ * `makeRetryService`) and threads it into `makeAPI({ retry: <SDK>.Retry })`
+ * so that callers can install a blanket retry policy at the layer level
+ * for that SDK without wrapping every call with `Effect.retry`.
  */
 import * as Duration from "effect/Duration";
 import * as Effect from "effect/Effect";
 import { pipe } from "effect/Function";
-import * as Layer from "effect/Layer";
 import * as Ref from "effect/Ref";
 import * as Schedule from "effect/Schedule";
 import * as Context from "effect/Context";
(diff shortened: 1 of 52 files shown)


packages/core/src/retry.ts
```diff
@@ -85,6 +85,49 @@ export const capped = (max: Duration.Duration) =>
...
+// ============================================================================
+// Server Hint Helpers
+// ============================================================================
+
+/**
+ * Hard cap on how long we'll honor a server-provided `Retry-After` /
+ * `RateLimit` hint. A misbehaving server can otherwise park us forever.
+ */
+const SERVER_HINT_CAP_MS = 60_000;
+
```
Mkassabov, line 155: hard coding this is vile. shove it in config/layer/env var

scripts/error-discovery.ts
```diff
@@ -112,6 +120,25 @@ Different SDKs use different patch formats:
...
+\`\`\`
+When you construct a retryable error (TooManyRequests/429, InternalServerError/500,
+BadGateway/502, ServiceUnavailable/503, GatewayTimeout/504, Locked/423, or any
+class categorized retryable), you MUST pass \`retryAfter: parseRetryAfterForStatus(status, headers)\`
+so the retry policy honors server-provided wait hints (Retry-After / RateLimit headers).
+Import: \`import { parseRetryAfterForStatus } from "@distilled.cloud/core/retry-after";\`
+If this service uses bespoke headers or body fields for cooldown, parse them in
+\`matchError\` and pass the resulting \`Duration\` as \`retryAfter\` instead of (or
+in addition to, with \`??\` fallback) \`parseRetryAfterForStatus(status, headers)\`.
+
```
Mkassabov, line 145: you should be able to make a retryable error without retryAfter and we assume a retry-immediately then exponential backoff model by default imo

PR #230, opened 2026-05-04
fix(planetscale): add missing nullable and optional fields to branch and role

lucasthevenet:
This PR fixes several fields from the branch and role schemas that are sometimes not returned or returned null.

The fields that were fixed are the following:

- In **DatabaseBranch**, the following fields are not returned when the related **Database** is **Postgres** instead of **MySQL**: "mysql_address", "mysql_edge_address", "direct_vtgate", "vtgate_size", "vtgate_count", "schema_ready", "sharded", "shard_count", "vtgate_options".
- In **PostgresRole**, password field is only returned when the Role is created, subsequent fetches return password: null.
- Additionally several fields in **PostgresRole** are nullable "ttl", "deleted_at", "expires_at", "dropped_at", "disabled_at", "drop_failed"


sam-goodwin:
Is their spec wrong or was our code generator wrong?

Mkassabov:
> Is their spec wrong or was our code generator wrong?

if these patches are correct their api is wrong

Mkassabov:
@lucasthevenet can you add test cases for these please.

sam-goodwin:
> > Is their spec wrong or was our code generator wrong?
> 
> if these patches are correct their api is wrong

Are all specs just wrong lol

lucasthevenet:
> @lucasthevenet can you add test cases for these please.

will do

lucasthevenet:
@Mkassabov just added some tests, Planetscale takes eons to get Postgres db's ready so I had to extend timeouts for some of them.

sam-goodwin:
@Mkassabov how is this looking?

PR #236, opened 2026-05-04
feat(polar): add Polar distilled SDK

carlulsoe:
## Summary

Adds `@distilled.cloud/polar`, an Effect-native SDK generated from Polar's OpenAPI spec. The package includes Polar credentials, client/error/retry wiring, generated operations, package README, CI/release/preview-package wiring, and focused sandbox tests for Products, Webhook Endpoints, Benefits, Checkout Links, Organization Access Tokens, Customers, Customer State, Custom Fields, Discounts, Meters, Events/Event Types, Files, Payments, and CSV Exports.

This also improves shared OpenAPI generation for Polar-shaped specs: top-level `oneOf` / `anyOf` request bodies now emit usable input schemas, OpenAPI `const` values become schema literals, simple nullable unions become `Schema.NullOr(...)`, and bare `token` response fields are treated as sensitive. OAuth dynamic client registration responses are patched from Polar's live response shape so `client_secret` and `registration_access_token` are typed and redacted instead of `Schema.Unknown`.

Polar resource output patches now flatten common fields for high-value discriminated outputs that otherwise generated as `Schema.Unknown`: Benefits, Customers, Customer State, Custom Fields, Discounts, Meters, Events, Files, Payments, CSV Exports, OAuth authorize/userinfo, customer-portal benefit grants/payment methods, and customer-portal payment-method create/confirm responses. This keeps the generated SDK useful without asking the shared generator to expand arbitrary recursive output unions.

## Notes

This was started with `create-sdk-full`. The automated Claude refinement step could not run locally because this account does not have Claude access, so the hand-written SDK wiring and spec patches were completed manually.

I also could not create `alchemy-run/distilled-spec-polar` from this account, so the PR currently vendors Polar's public OpenAPI JSON and `specs:fetch` downloads directly from `https://api.polar.sh/openapi.json`. Happy to switch this to the normal `alchemy-run/distilled-spec-polar` submodule if you create that mirror repo.

OAuth client CRUD has schema/redaction coverage, but not live SDK CRUD coverage yet: Polar's create endpoint returns a registration access token that must be used for subsequent get/update/delete, while this SDK currently uses one configured bearer token per client. OAuth authorize/userinfo also have schema coverage only because they require browser/OIDC session tokens rather than the organization access token used by the live sandbox suite.

Customer-portal benefit grants, payment methods, and payment-method create/confirm responses have schema coverage, but not live coverage; those endpoints require customer/member session auth instead of the organization access token used by the live sandbox suite.

## Validation

- `bun run generate` in `packages/polar`
- `bun run check` in `packages/polar`
- `bun run build` in `packages/polar`
- `bun run test` in `packages/polar`: 26 passed, 30 live tests skipped
- `POLAR_SERVER=sandbox bun run test` in `packages/polar` with live sandbox token: 56 passed
- secret scan for persisted Polar/GitHub tokens: no matches
- `bun run check` in `packages/core`
- `bun run test` in `packages/core`: 35 passed

## Follow-ups

- Decide whether OAuth registration-token management should be modeled as per-call auth override in the SDK or left to the Alchemy provider layer.
- Decide whether empty-body 200 responses such as OAuth client delete should be modeled as a no-content output instead of preserving `Schema.Unknown`.
- Build the Alchemy IaC provider on top of the SDK after this foundation lands.


Mkassabov:
clankers make bad comments.

I think this is done but tests are hitting rate limits

Mkassabov:
done https://github.com/alchemy-run/distilled/pull/484

PR #257, opened 2026-05-05
fix(cloudflare/rulesets): support phase entrypoint paths

agcty:
Fixes generated Cloudflare Rulesets phase-entrypoint requests so the SDK can call endpoints such as `/zones/{zone_id}/rulesets/phases/{rulesetPhase}/entrypoint`.

- synthesizes missing URL path params such as `rulesetPhase` during generation
- keeps optional generated path params optional in request schemas
- maps Cloudflare `{accountOrZone}/{accountOrZoneId}` paths from exactly one of `accountId` or `zoneId`
- adds request-building coverage for zone-scoped Rulesets entrypoint updates

Verification run:

- `bun --filter '@distilled.cloud/cloudflare' test --run test/client-api.test.ts`
- `bun --filter '@distilled.cloud/cloudflare' typecheck`
- `bun --filter '@distilled.cloud/cloudflare' check`
- `git diff --check`

### DIFF
--- a/packages/cloudflare/src/client/api.ts
+++ b/packages/cloudflare/src/client/api.ts
@@ -38,7 +38,12 @@ import {
 } from "../errors.ts";
 import { Credentials, formatHeaders } from "../credentials.ts";
 import { Retry } from "../retry.ts";
-import { type ErrorMatcher, getErrorMatchers } from "../traits.ts";
+import {
+  type AccountOrZoneScopeTrait,
+  type ErrorMatcher,
+  getAccountOrZoneScope,
+  getErrorMatchers,
+} from "../traits.ts";
 
 // ============================================================================
 // Global Cloudflare error codes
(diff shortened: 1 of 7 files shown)


agcty:
Verification update:

- Focused test passed: `bun --filter @distilled.cloud/cloudflare test --run test/client-api.test.ts`.
- Typecheck passed: `bun --filter @distilled.cloud/cloudflare typecheck`.
- Check passed: `bun --filter @distilled.cloud/cloudflare check`.

This fixes the generated Rulesets phase-entrypoint path used by alchemy-run/alchemy-effect#240. Without this fix, the Alchemy live test calls `/{accountOrZone}/{accountOrZoneId}/rulesets/phases/...` through the generated client and Cloudflare returns `No route for that URI`.

packages/cloudflare/src/client/api.ts
```diff
@@ -402,25 +402,56 @@ class CloudflareDecodeError extends CloudflareHttpError {
...
+    next = {
+      ...next,
+      path: next.path
+        .replace("{accountOrZone}", hasAccountId ? "accounts" : "zones")
+        .replace(
+          "{accountOrZoneId}",
+          encodeURIComponent(String(hasAccountId ? accountId : zoneId)),
+        ),
+    };
+  }
```
sam-goodwin: @Mkassabov is this something we can handle with traits? Looks to be getting a bit unwieldy?

Mkassabov: I think so? I'd rather not ship logic like this in the api layer if we can avoid it.

agcty: updated this to use a Cloudflare-specific `T.AccountOrZoneScope()` trait instead of inferring it from the path in the api layer. core now passes the input schema to request transforms, the generator emits the trait for account-or-zone scoped operations, and the focused client-api test covers the rulesets path mapping. checks passed: `bun test packages/cloudflare/test/client-api.test.ts`, `bun --filter @distilled.cloud/core typecheck`, `bun --filter @distilled.cloud/cloudflare typecheck`, `bun --filter @distilled.cloud/core lint`, `bun --filter @distilled.cloud/cloudflare lint`.

Mkassabov:
@agcty can you take a look at PR #282 does that solve your issues?

agcty:
@Mkassabov yes i believe it does, will close this PR and adapt the alchemy-effect one based on 282

agcty:
@Mkassabov can test once distilled is released again

PR #261, opened 2026-05-05
fix(core,gcp): honor T.Service.rootUrl and preserve {+name} reserved-expansion

microagi-andy:
After #260 every read-by-name request from `@distilled.cloud/gcp` still fails — but with a different error than before. Two distinct gaps surface, and they share enough structure to fix as one PR.

```
error: InvalidUrl error (GET v3/projects%2Falchemy-test-rpilvq)
       _tag: "HttpClientError"
  request: { method: "GET", url: "v3/projects%2Falchemy-test-rpilvq", ... }
```

Both failure signatures are visible in that one URL: `v3/projects%2F...` (slash percent-encoded) and no `https://cloudresourcemanager.googleapis.com/` prefix.

### Gap 1 — `T.Service.rootUrl` is never applied

`packages/core/src/client.ts` only consulted `config.getBaseUrl(creds)`. GCP's `makeAPI` config returns `""` (hosts vary per service, not per credentials) and instead declares the host on the service trait:

```ts
// packages/gcp/src/services/cloudresourcemanager-v3.ts
const svc = T.Service({
  name: "cloudresourcemanager",
  version: "v3",
  rootUrl: "https://cloudresourcemanager.googleapis.com/",
  servicePath: "",
});
```

The trait shape (`packages/core/src/traits.ts:384`) and the accessor (`getServiceTrait`) already exist — `client.ts` just never read them. Result: the resolved URL was a bare relative path and `effect/unstable/http/HttpClient` rejected it as `ERR_INVALID_URL`.

### Gap 2 — `{+name}` reserved-expansion regressed to `{name}`

#260 fixed flatPath/path mismatches by stripping `{+name}` to `{name}` in the generator so a single `path.replace(\`{${name}}\`, encodeURIComponent(value))` call would substitute. But `encodeURIComponent` percent-encodes `/` — and GCP discovery documents use `{+name}` precisely because the value (`projects/foo`) contains `/` that must stay literal under RFC 6570 §3.2.3 reserved-expansion.

#260's regression test masked this:

```ts
expect([
  "v3/projects/microagi-research",
  "v3/projects%2Fmicroagi-research",
]).toContain(parts.path);
```

Either form passed; the implementation produced the second; nobody noticed.

### Why one PR

The two fixes compose: stop stripping `{+name}` (generator), recognise it (core), and prepend the per-service host (core). Each piece is no use without the other two on the GCP path.

### Changes

- `core/src/traits.ts`: `buildRequestParts` recognises `{+name}` and substitutes via a new `encodeReserved` helper that preserves RFC 3986 reserved + unreserved characters; falls back to `encodeURIComponent` for plain `{name}`. Pure addition — simple expansion is unchanged.
- `core/src/client.ts`: when `config.getBaseUrl(creds)` is empty, fall back to `Traits.getServiceTrait(inputSchema.ast).rootUrl + servicePath`. Single declaration; both the request-build site and the paginated branch share it.
- `gcp/scripts/generate.ts`: emit `method.path` verbatim. Drop `stripReservedExpansion`.
- `gcp/src/services/*.ts`: regenerated. Mechanical churn — every operation that addresses a resource by name moves from `path: "v3/{name}"` to `path: "v3/{+name}"` (~417 files).
- `gcp/src/client/api.ts`: tighten the misleading "via Http trait" comment ("via Service trait"). Behaviour unchanged.

```diff
-const baseUrl = config.getBaseUrl(creds as ResolvedCreds);
+let baseUrl = config.getBaseUrl(creds as ResolvedCreds);
+if (!baseUrl) {
+  const svcTrait = Traits.getServiceTrait(inputSchema.ast);
+  if (svcTrait?.rootUrl) {
+    baseUrl = svcTrait.rootUrl + (svcTrait.servicePath ?? "");
+  }
+}
```

### Tests

- `packages/core/test/build-request-parts.test.ts` (new): pins `{+name}` preserves `/`, `{name}` encodes `/`, `{+name}` still encodes whitespace, and a fully-reserved value round-trips literally. Synthetic schemas — no GCP dependency.
- `packages/gcp/test/path-substitution.test.ts` (rewritten): drops the permissive `expect([..., "%2F..."]).toContain(parts.path)` form. Asserts the literal-`/` form exactly on real `getProjects` / `deleteProjects`. Adds a whitespace case.
- `packages/gcp/test/base-url.test.ts` (new): mocks `HttpClient.HttpClient`, captures the outgoing request URL, and asserts `getProjects` / `deleteProjects` resolve to `https://cloudresourcemanager.googleapis.com/v3/projects/<name>`. End-to-end regression for both gaps without a network call.

### Why other distilled SDKs are unaffected

- AWS bypasses `core.makeAPI` entirely (per-region/per-service endpoint resolution via Smithy rules).
- Cloudflare's `getBaseUrl(creds)` returns `creds.apiBaseUrl` — non-empty — so the rootUrl fallback never runs. `{+name}` doesn't appear in Cloudflare's specs.
- Verified locally: typecheck clean for `aws`, `cloudflare`, `neon`, `planetscale`, `axiom`, `stripe`, `supabase`, `fly-io`, `coinbase` against the modified `core`.

### Verification

```
bun --filter '@distilled.cloud/core' run test  →  39 passed
bun --filter '@distilled.cloud/gcp'  run test  →   5 passed
bun --filter '@distilled.cloud/{core,gcp}' run check  →  clean
```

End-to-end against live GCP using a downstream consumer (`alchemy-v2-gcp`):

```
cd vendor/alchemy-v2-gcp && CI=1 bun test test/CloudResourceManager/Project.test.ts
```

### DIFF
--- a/packages/gcp/patches/container/createProjectsLocationsClustersNodePools.json
+++ b/packages/gcp/patches/container/createProjectsLocationsClustersNodePools.json
@@ -1,0 +1,7 @@
+{
+  "errors": {
+    "BadRequest": [{ "httpStatus": 400 }],
+    "Forbidden": [{ "httpStatus": 403 }],
+    "Conflict": [{ "httpStatus": 409 }]
+  }
+}
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -444,7 +444,15 @@ export const makeAPI = <Creds>(config: ClientConfig<Creds>) => {
             : credentials;
           const client = yield* HttpClient.HttpClient;
 
-          const baseUrl = config.getBaseUrl(creds as ResolvedCreds);
+          // Fall back to the Service trait when the consumer leaves
+          // `getBaseUrl` empty (per-service hosts rather than per-credentials).
+          let baseUrl = config.getBaseUrl(creds as ResolvedCreds);
+          if (!baseUrl) {
+            const svcTrait = Traits.getServiceTrait(inputSchema.ast);
+            if (svcTrait?.rootUrl) {
+              baseUrl = svcTrait.rootUrl + (svcTrait.servicePath ?? "");
+            }
+          }
           const authHeaders = config.getAuthHeaders(creds as ResolvedCreds);
 
           // Use schema-aware request builder for proper camelCase → wire_name mapping
--- a/packages/core/src/traits.ts
+++ b/packages/core/src/traits.ts
@@ -623,6 +623,14 @@ const unwrapRedactedDeep = (value: unknown): unknown => {
   return value;
 };
 
+/**
+ * RFC 6570 §3.2.3 reserved-expansion: encode everything outside the RFC 3986
+ * unreserved (`A-Za-z0-9-._~`) and reserved (`:/?#[]@!$&'()*+,;=`) sets.
+ */
+const RFC3986_NEEDS_ENCODING = /[^A-Za-z0-9\-._~:/?#\[\]@!$&'()*+,;=]/g;
+const encodeReserved = (v: string): string =>
+  v.replace(RFC3986_NEEDS_ENCODING, encodeURIComponent);
+
 export const buildRequestParts = (
   ast: AST.AST,
   httpTrait: HttpTrait,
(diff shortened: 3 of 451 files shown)


Mkassabov:
since you run core here can you run generate in the root. I want to see if this triggers changes in any of the other SDKs, since I'd have to look closer at this if it does

packages/gcp/patches/container/createProjectsLocationsClustersNodePools.json
```diff
@@ -0,0 +1,7 @@
+{
+  "errors": {
+    "BadRequest": [{ "httpStatus": 400 }],
+    "Forbidden": [{ "httpStatus": 403 }],
+    "Conflict": [{ "httpStatus": 409 }]
+  }
+}
```
sam-goodwin, line 7: These look to be all the same everywhere. Should we instead encode that in the code generator scrpt instead of a ton of patch files? Or is it not universal?

microagi-andy:
My bad - I only wanted the first 3 commits in this PR. Closing this in favour https://github.com/alchemy-run/distilled/pull/262

PR #275, opened 2026-05-07
feat(neon): bump spec submodule to 2026-05-07

sam-goodwin:
Bumps the Neon spec submodule to 2026-05-07. Picks up upstream additions including the `OperationAction.timeline_update_protected_config` literal that the live API returns after toggling `branch.protected` — without it, `getProjectOperation` fails to decode and any caller polling operations breaks.

```
SchemaError: Expected "create_compute" | ... | "swap_binding_id", got "timeline_update_protected_config"
```

Spec `1695ca9 → 209b520`, then regenerate.


packages/neon/src/operations/addProjectJWKS.ts
```diff
@@ -59,10 +59,13 @@ export const AddProjectJWKSOutput = /*@__PURE__*/ /*#__PURE__*/ Schema.Struct({
         "sync_dbs_and_roles_from_compute",
         "apply_schema_from_branch",
         "timeline_mark_invisible",
+        "timeline_update_protected_config",
         "prewarm_replica",
         "promote_replica",
         "set_storage_non_dirty",
         "swap_binding_id",
+        "finalize_migration",
+        "mark_migration_prepared",
```
sam-goodwin, line 68: @Mkassabov apis changing should not break our schemas unless they are breaking changes. 

Adding items to a union is not breaking, but it breaks distilled because we are too strict. 

Should we open it up somehow?

PR #282, opened 2026-05-07
fix(cloudflare): split account-or-zone scoped operations into per-scope variants

Mkassabov:
Cloudflare's API has a path pattern `/{accountOrZone}/{accountOrZoneId}/...` where the literal first segment is `accounts` or `zones` depending on which scope ID the caller provides. The OpenAPI spec collapses both into one entry that declares `account_id` and `zone_id` as both required path params (mutually exclusive in reality). Distilled's generator was faithfully turning this into broken code: both fields typed as required, and `{accountOrZone}` / `{accountOrZoneId}` placeholders never bound to anything, so ~85 endpoints across 7 services were uncallable.

This PR fixes those endpoints by splitting each at codegen time into two concrete operations plus a thin convenience wrapper. Supersedes #257, which used a runtime trait + `throw new Error` for the same problem.

### Generator changes

Three independent fixes in `packages/cloudflare/scripts/generate.ts`:

1. **Auto-detect `accountOrZone`-scoped operations** by URL template and split each into:
   - `<op>ForAccount` — concrete URL `/accounts/{account_id}/...`, request schema with `accountId`
   - `<op>ForZone` — concrete URL `/zones/{zone_id}/...`, request schema with `zoneId`
   - `<op>` — wrapper function dispatching on `"accountId" in input`. Returns `Effect.Effect<...>` (or `stream.Stream<...>` for paginated). Input is the discriminated union, so invalid states are unrepresentable at the type level.
   - Body fields are extracted into a non-exported `<Op>BaseFields` const and spread into both variants — no source-level duplication.
   - Response and error types remain singular and shared between variants.

2. **Synthesize missing URL path params**. Some operations reference placeholders like `{rulesetPhase}` in the URL but don't declare them as parameters — these are now auto-synthesized as required string path params.

3. **Optional path params** now correctly emit as `Schema.optional(...)` instead of being typed as required.

### User-visible API

Before (broken — no way to substitute `accountOrZone`/`accountOrZoneId`, both ids typed required):

```ts
yield* cf.rulesets.putPhas({ accountId, zoneId, rulesetPhase: "...", rules: [...] });
// → PUT /{accountOrZone}/{accountOrZoneId}/rulesets/phases/.../entrypoint  (404)
```

After:

```ts
// Single function, exactly-one-of input enforced by the type system
yield* cf.rulesets.putPhas({ accountId: "...", rulesetPhase: "...", rules: [...] });
yield* cf.rulesets.putPhas({ zoneId:    "...", rulesetPhase: "...", rules: [...] });

// Or import the concrete variant directly for tighter tree-shaking
yield* cf.rulesets.putPhasForAccount({ accountId: "...", rulesetPhase: "...", rules: [...] });
```

### Why split at codegen instead of runtime

Considered a runtime approach (PR #257): mark the schema with a trait, rewrite the path in `transformRequestParts`, throw on invalid input. Rejected because (a) the throw bypasses Effect's typed error channel — it becomes an unchecked defect rather than a typed operation error, (b) "exactly one of" is encodable in the type system and shouldn't be a runtime check, and (c) it imports Cloudflare-specific path-rewriting logic into the core client. Splitting at codegen keeps core untouched, errors flow through Effect normally, and the type system enforces the invariant.

### Test coverage

Added in `packages/cloudflare/test/client-api.test.ts`:
- `putPhasForAccount` builds `/accounts/{id}/...` correctly
- `putPhasForZone` builds `/zones/{id}/...` correctly
- Wrapper dispatches to account variant when `accountId` is present
- Wrapper dispatches to zone variant when `zoneId` is present
- Source-level smoke check that the wrapper uses `"accountId" in input`

All 7 client-api tests pass; `typecheck` and `check` are clean for both the cloudflare package and core.

### Out of scope (separate issues to file)

- Truncated upstream operationIds (`getPhas`, `putPhas`, `listPhasVersions`, etc.) — these come from Cloudflare's own published OpenAPI spec and need a spec patch, not a generator fix.
- Refactoring core's `buildRequestParts` to handle `Schema.Union` natively (would let SDKs express discriminated request unions generically without codegen-time splitting).

### DIFF
--- a/packages/cloudflare/src/services/firewall.ts
+++ b/packages/cloudflare/src/services/firewall.ts
@@ -5,6 +5,8 @@
  * DO NOT EDIT - regenerate with: bun scripts/generate.ts --service firewall
  */
 
+import * as Effect from "effect/Effect";
+import * as stream from "effect/Stream";
 import * as Schema from "effect/Schema";
 import type * as HttpClient from "effect/unstable/http/HttpClient";
 import * as API from "../client/api.ts";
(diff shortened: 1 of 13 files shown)


packages/cloudflare/src/services/firewall.ts
```diff
@@ -346,11 +448,51 @@ export const listAccessRules: API.PaginatedOperationMethod<
...
+export const listAccessRules = (
+  input: ListAccessRulesRequest,
+): stream.Stream<
+  ListAccessRulesResponse,
+  ListAccessRulesError,
+  Credentials | HttpClient.HttpClient
+> =>
+  "accountId" in input
+    ? listAccessRulesForAccount.pages(input)
+    : listAccessRulesForZone.pages(input);
```
sam-goodwin: should probably be an error if both are provided? ambiguity?

sam-goodwin: why are you calling .pages here?

what about `yield* listAccessRules` .items, .pages

PR #285, opened 2026-05-08
feat(core/openapi): emit makePaginated for cursor/page/token list ops

sam-goodwin:
Today every generated list op is `API.make(...)` and callers have to roll their own `do { … } while (cursor)` loop. After this change the generator detects pagination and emits `API.makePaginated(...)` so every list op gets `.pages()` / `.items()` Stream methods for free:

```ts
// before — manual cursor loop
let cursor: string | undefined;
do {
  const page = yield* listProjectBranches({ project_id, search: name, ...(cursor ? { cursor } : {}) });
  for (const b of page.branches) if (b.name === name) matches.push(b);
  cursor = page.pagination?.next;
} while (cursor);

// after — yield* the items stream
yield* listProjectBranches.items({ project_id, search: name }).pipe(
  Stream.filter((b) => b.name === name),
  Stream.runCollect,
);
```

### Detection rules

Inspecting the resolved input parameters and response schema:

| output marker | mode |
|---|---|
| `pagination.cursor` / `pagination.next` | `cursor` |
| `pagination.next_page` / top-level `next_page` | `page` |
| top-level `next_token` / `NextToken` | `token` |

When a marker is found, the generator looks for a matching input parameter (`cursor` / `page_token` / `next_token` / `page`) and the first non-`pagination` top-level array property as the items path. If all three line up, emit `makePaginated`; otherwise fall back to `make`.

### Impact

50 ops across `@distilled.cloud/{neon,planetscale,stripe}` flip from `make` to `makePaginated`. Type-check is clean across the workspace.

### DIFF
--- a/packages/core/scripts/generate-openapi.ts
+++ b/packages/core/scripts/generate-openapi.ts
@@ -1232,6 +1232,12 @@ export function generateFromOpenAPI(config: GeneratorConfig): void {
             }
           }
 
+          const pagination = detectPagination(
+            parameters as ParameterObject3[],
+            responseSchema,
+            spec,
+          );
+
           const code = buildOperationFile(
             functionName,
             inputSchemaCode,
@@ -1241,6 +1247,7 @@ export function generateFromOpenAPI(config: GeneratorConfig): void {
             jsDoc,
             operationErrors,
             sensitiveImports,
+            pagination,
             config,
           );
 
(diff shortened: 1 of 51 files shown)


packages/core/scripts/generate-openapi.ts
```diff
@@ -1394,6 +1408,117 @@ export function generateFromOpenAPI(config: GeneratorConfig): void {
...
 // ============================================================================
 
+/**
+ * Detect cursor / page / token pagination on an operation by looking at the
+ * input parameters and the (resolved) response schema.
+ *
+ * Returns the `pagination` trait to feed into `API.makePaginated`, or
+ * `undefined` if the operation isn't paginated.
+ */
+function detectPagination(
```
sam-goodwin, line 1418: this looks concerning, maybe?

PR #294, opened 2026-05-13
fix(planetscale): Add missing error return codes and nullable fields to branch, resize request, roles

lucasthevenet:
Here we go again.. Planetscale's API is making me have trust issues lol.

This PR addresses the following:

- **deleteBranch** can return 422 if trying to delete a db's default branch
- **deleteRole** can return 422 if the role is still referenced. [See mention of this on Planetscale Docs](https://planetscale.com/docs/postgres/connecting/roles#:~:text=If%20you%20try%20to%20delete%20a%20role%20that%20is%20still%20referenced%2C%20you%20may%20see%20this%20error:%20Role%20is%20still%20referenced%20and%20cannot%20be%20dropped..)
- **PostgresClusterResizeRequest** can return completed_at = null if the operation has not finished yet
- Referenced **branches** can return deleted_at = null if not deleted

### DIFF
--- a/packages/planetscale/tests/branches.test.ts
+++ b/packages/planetscale/tests/branches.test.ts
@@ -1,7 +1,7 @@
 import { Effect, Schedule } from "effect";
 import { afterAll, beforeAll, describe, expect, it } from "vitest";
 import { UnknownPlanetScaleError } from "../src/client";
-import { Forbidden, NotFound } from "../src/errors";
+import { Forbidden, NotFound, UnprocessableEntity } from "../src/errors";
 import { createBranch } from "../src/operations/createBranch";
 import { deleteBranch } from "../src/operations/deleteBranch";
 import { demoteBranch } from "../src/operations/demoteBranch";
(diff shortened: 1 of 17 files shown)


packages/planetscale/tests/branches.test.ts
```diff
@@ -536,54 +537,58 @@ describe.each([{ kind: "postgresql" }, { kind: "mysql" }] as const)(
...
+
+          // Create branch
+          const created = await runEffect(
+            createBranch({
+              organization: db.organization,
+              database: db.name,
+              name: branchName,
+              parent_branch: "main",
+            }),
+          );
```
sam-goodwin, line 555: Hmm, this was before your change but we should use the effect vitest wrapper so we can do this with native effect instead of async code. runEffect is a smell.

sam-goodwin (approved):

sam-goodwin:
@Mkassabov does this look good to merge ?

Mkassabov (approved):

PR #303, opened 2026-05-20
fix(cloudflare): tag Turnstile missing widgets as NotFound

agcty:
## Summary
- map Turnstile missing-widget 404 responses to a typed `NotFound` error
- let per-operation error matchers run for non-JSON and non-envelope Cloudflare errors
- add regression tests for the raw 404 shapes that showed up while testing the alchemy-effect Turnstile resource

This unblocks cleaning up alchemy-effect PR #370 so it can catch `NotFound` instead of checking for `CloudflareHttpError` + status 404.

## Verification
- `bun --filter @distilled.cloud/cloudflare test -- test/client-api.test.ts`
- `bun --filter @distilled.cloud/cloudflare typecheck`
- `bun --filter @distilled.cloud/cloudflare check`

### DIFF
--- a/packages/cloudflare/src/client/api.ts
+++ b/packages/cloudflare/src/client/api.ts
@@ -256,6 +256,10 @@ function findMatchingError(
   return bestMatch;
 }
 
+// Some Cloudflare endpoints return bare HTTP failures instead of the standard
+// `{ success: false, errors: [...] }` envelope. Generated operation schemas can
+// still declare status-based error matchers, so try those before falling back
+// to generic HTTP errors.
 function matchOperationError(
   errors: readonly ApiErrorClass[] | undefined,
   code: number | undefined,
--- a/packages/cloudflare/src/client/api.ts
+++ b/packages/cloudflare/src/client/api.ts
@@ -220,6 +220,12 @@ function matcherSpecificity(matcher: ErrorMatcher): number {
   return score;
 }
 
+interface OperationErrorSchema {
+  schema: Schema.Top;
+  tag: string;
+  matchers: readonly ErrorMatcher[];
+}
+
 interface MatchedError {
   schema: Schema.Top;
   tag: string;
@@ -229,7 +235,7 @@ interface MatchedError {
  * Find matching error schema using annotations from the schema AST.
  */
 function findMatchingError(
-  errorSchemas: Map<string, Schema.Top>,
+  errorSchemas: readonly OperationErrorSchema[],
   code: number | undefined,
   status: number,
   message: string,
@@ -237,17 +243,16 @@ function findMatchingError(
   let bestMatch: MatchedError | undefined;
   let bestScore = 0;
 
-  for (const [name, schema] of errorSchemas) {
-    const ast = schema.ast;
-    const matchers = getErrorMatchers(ast);
-    if (!matchers || matchers.length === 0) continue;
-
-    for (const matcher of matchers) {
+  for (const errorSchema of errorSchemas) {
+    for (const matcher of errorSchema.matchers) {
       if (matchesExpression(matcher, code, status, message)) {
         const score = matcherSpecificity(matcher);
         if (score > bestScore) {
           bestScore = score;
-          bestMatch = { schema, tag: name };
+          bestMatch = {
+            schema: errorSchema.schema,
+            tag: errorSchema.tag,
+          };
         }
       }
     }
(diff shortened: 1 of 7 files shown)


packages/cloudflare/src/services/turnstile.ts
```diff
@@ -12,6 +12,17 @@ import * as T from "../traits.ts";
...
+// =============================================================================
+// Errors
+// =============================================================================
+
+export class NotFound extends Schema.TaggedErrorClass<NotFound>()("NotFound", {
+  message: Schema.String,
+}) {}
+// Turnstile returns a bare 404 for missing widgets instead of Cloudflare's
+// normal error envelope, so match the HTTP status directly.
+T.applyErrorMatchers(NotFound, [{ status: 404 }]);
```
Mkassabov, line 23: these are generated files do not modify them directly. add a patch and then rerun generate

packages/cloudflare/src/client/api.ts
Mkassabov, line 1: why are we making changes to this file?

agcty: this change is needed because Turnstile returns bare/non-envelope 404 responses for missing widgets. the generated operation patch can now declare `{ status: 404 }`, but `api.ts` has to try operation-specific matchers before falling back to generic `CloudflareHttpError` for those non-envelope responses. added comments and regression tests for both non-json and non-envelope cases.

packages/cloudflare/src/services/turnstile.ts
agcty: fixed: moved this into `packages/cloudflare/patches/turnstile/{getWidget,deleteWidget}.json` and reran `bun scripts/generate.ts --service turnstile`, so `turnstile.ts` is generated again.

packages/cloudflare/src/client/api.ts
```diff
@@ -256,10 +256,55 @@ function findMatchingError(
   return bestMatch;
 }
 
+function matchOperationError(
```
sam-goodwin: Hmm, why is this necessary?

Mkassabov: it seems to be to handle when there is no envelope from cloudflare; I found it weird when i first looked at this, but it seems fine

agcty: Added a `//` comment above `matchOperationError` explaining the Cloudflare no-envelope case. The generated Turnstile operation patch declares the 404 matcher, but `api.ts` has to try operation-specific matchers before falling back to `CloudflareHttpError` when Cloudflare returns a bare HTTP failure. Re-ran `bun --filter @distilled.cloud/cloudflare check` and `bun run vitest run test/client-api.test.ts` from `packages/cloudflare`.

packages/cloudflare/src/client/api.ts
```diff
@@ -256,10 +256,59 @@ function findMatchingError(
...
+
+  const errorSchemas = new Map<string, Schema.Top>();
+  for (const errorSchema of errors) {
+    const identifier = extractTagFromAst(
+      (errorSchema as unknown as Schema.Top).ast,
+    );
+    if (identifier) {
+      errorSchemas.set(identifier, errorSchema as unknown as Schema.Top);
+    }
+  }
```
sam-goodwin: We shouldn't do all this work on every request. Instead, it should create this map and return a function for matching, that way we cache the map in the closure once.

agcty: addressed, lmk what you think

packages/cloudflare/src/client/api.ts
```diff
@@ -256,22 +261,95 @@ function findMatchingError(
...
  * Match a Cloudflare API error response using per-operation error schemas.
  */
-const matchError = (
+export const matchCloudflareError = (
   status: number,
   errorBody: unknown,
   errors?: readonly ApiErrorClass[],
   headers?: Record<string, string | undefined>,
 ): Effect.Effect<never, unknown> => {
+  const matchOperationError = getOperationErrorMatcher(errors);
```
sam-goodwin, line 338: This is not caching anything. It bulds the matcher on every request still

packages/cloudflare/src/client/api.ts
```diff
@@ -256,22 +261,95 @@ function findMatchingError(
...
+// `{ success: false, errors: [...] }` envelope. Generated operation schemas can
+// still declare status-based error matchers, so try those before falling back
+// to generic HTTP errors.
+function getOperationErrorMatcher(
+  errors: readonly ApiErrorClass[] | undefined,
+): OperationErrorMatcher {
+  if (!errors || errors.length === 0) return noOperationErrorMatcher;
+
+  const cached = operationErrorMatcherCache.get(errors);
+  if (cached) return cached;
```
sam-goodwin, line 286: Using a global map is a massive code smell

sam-goodwin: I looked into, i can see why AI is doing that. API.make's `matchError` function is limiting. We really need to rework this part of the codebase.

Mkassabov:
closing in favor of #402; the distilled v1 rewrite handled a lot of the code logic so this could be handled by just a patch file

@agcty thanks for flagging the issue!

PR #304, opened 2026-05-22
fix(cloudflare): add missing binding types to workers edge preview api

john-royal:
The edge preview create route was missing a few binding types. This is blocking `cloudflare-runtime` from supporting them as remote bindings.

I used the `WorkerMetadataBinding` type from workers-sdk as a reference ([link](https://github.com/cloudflare/workers-sdk/blob/07336888e0bc82925e4023f5b72a0062f10d77b8/packages/workers-utils/src/types.ts#L56)), which is what Wrangler does.

### DIFF
--- a/packages/cloudflare/specs/cloudflare/workers.openapi.yml
+++ b/packages/cloudflare/specs/cloudflare/workers.openapi.yml
@@ -8700,6 +8700,8 @@ paths:
                                 type: string
                               id:
                                 type: string
+                              internalEnv:
+                                type: string
                               raw:
                                 type: boolean
                             required:
@@ -8737,6 +8739,8 @@ paths:
                               environment:
                                 type: string
                               entrypoint:
+                                type: string
+                              cross_account_grant:
                                 type: string
                             required:
                               - type
@@ -8792,6 +8796,8 @@ paths:
                               name:
                                 type: string
                               index_name:
+                                type: string
+                              internalEnv:
                                 type: string
                               raw:
                                 type: boolean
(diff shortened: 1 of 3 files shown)


packages/cloudflare/specs/cloudflare/workers.openapi.yml
```diff
@@ -9113,6 +9119,76 @@ paths:
...
+                                enum:
+                                  - unsafe_hello_world
+                              name:
+                                type: string
+                              enable_timer:
+                                type: boolean
+                            required:
+                              - type
+                              - name
+                          - type: object
```
sam-goodwin, line 9149: This looks like it's modifying the generated openapi spec directly instead of a patch?

john-royal: Not quite.

I updated the patch file at `packages/cloudflare/specs/cloudflare-patch/edge-preview.openapi.yml`. Everything else is generated, including the updated spec.

sam-goodwin: Seems like there are more changes than expected?

john-royal: Did you think this was the durable object change? This is separate.

This is adding missing binding types to the edge preview API, so we can support them as remote bindings in cloudflare-runtime. There were a few.

I just compared the patch and spec side by side and it doesn't look like anything was added beyond what's in the patch.

PR #306, opened 2026-05-25
fix(aws): strip "Message" suffix when deriving operation name

zeyuri:
AWS query-protocol services in the RDS family (RDS, ElastiCache, Redshift, Neptune, DocDB, …) name their input shapes `XxxMessage` (e.g. `DescribeDBInstancesMessage`, `CreateDBInstanceMessage`) rather than `XxxRequest`. The operation-name derivation only stripped a `Request`/`Input` suffix, so the aws-query serializer emitted `Action=DescribeDBInstancesMessage` and AWS rejected the call:

```
Could not find operation DescribeDBInstancesMessage for version 2014-10-31
```

This broke **every** RDS-family operation (describe and create) through the SDK. EC2 is unaffected because its inputs end in `…Request`.

### Fix

Strip `Message` in addition to `Request`/`Input` at the four sites that derive the op name from the input shape identifier:

```diff
-identifier.replace(/(?:Request|Input)$/, "")
+identifier.replace(/(?:Request|Input|Message)$/, "")
```

- `packages/aws/src/protocols/aws-query.ts` — `Action=` for RDS-family query services (the real fix)
- `packages/aws/src/protocols/ec2-query.ts` — `Action=` for EC2 query
- `packages/aws/src/protocols/aws-json.ts` — `X-Amz-Target` op name (JSON 1.0/1.1)
- `packages/aws/src/client/api.ts` — op name used for error context

Added a regression test asserting RDS `DescribeDBInstancesMessage` serializes to `Action=DescribeDBInstances` / `Version=2014-10-31`. It fails on the old regex and passes with the fix.

### Note for maintainers

This regex is a heuristic. A more robust approach would carry the operation's real shape-id / name from codegen instead of re-deriving it from the input identifier — happy to do that instead if you'd prefer. No AWS op's name itself ends in `Message` while its input is `<name>Message`, so stripping `Message` is safe against the smithy models.

### DIFF
--- a/packages/aws/src/protocols/aws-query.ts
+++ b/packages/aws/src/protocols/aws-query.ts
@@ -58,9 +58,13 @@ export const awsQueryProtocol: Protocol = (
   // Pre-compute encoder (done once at init)
   const encodeInput = S.encodeEffect(inputSchema);
 
-  // Pre-compute operation name and version from annotations
+  // Pre-compute operation name and version from annotations.
+  // RDS-family query services (RDS, ElastiCache, Redshift, Neptune, DocDB, …)
+  // name their input shapes "XxxMessage" (e.g. DescribeDBInstancesMessage), so
+  // we must strip "Message" in addition to "Request"/"Input" to recover the
+  // Action name.
   const identifier = getIdentifier(inputAst) ?? "";
-  const action = identifier.replace(/(?:Request|Input)$/, "");
+  const action = identifier.replace(/(?:Request|Input|Message)$/, "");
   const version = getServiceVersion(inputAst) ?? "";
 
   return {
--- a/packages/aws/src/client/api.ts
+++ b/packages/aws/src/client/api.ts
@@ -52,7 +52,7 @@ export const make = <Op extends Operation<any, any, any>>(
     // Extract metadata for error recording (DISTILLED_AWS_DEBUG) and error context
     const serviceSdkId = getAwsApiService(inputAst)?.sdkId;
     const operationName = getIdentifier(inputAst)?.replace(
-      /(?:Request|Input)$/,
+      /(?:Request|Input|Message)$/,
       "",
     );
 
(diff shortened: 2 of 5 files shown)


packages/aws/src/protocols/aws-query.ts
```diff
@@ -58,9 +58,13 @@ export const awsQueryProtocol: Protocol = (
...
 
-  // Pre-compute operation name and version from annotations
+  // Pre-compute operation name and version from annotations.
+  // RDS-family query services (RDS, ElastiCache, Redshift, Neptune, DocDB, …)
+  // name their input shapes "XxxMessage" (e.g. DescribeDBInstancesMessage), so
+  // we must strip "Message" in addition to "Request"/"Input" to recover the
+  // Action name.
   const identifier = getIdentifier(inputAst) ?? "";
-  const action = identifier.replace(/(?:Request|Input)$/, "");
+  const action = identifier.replace(/(?:Request|Input|Message)$/, "");
```
sam-goodwin, line 67: Is this something that should be fixed globally like this or applied by the code generator? Is it safe to always apply this rule?

zeyuri: Good question. I checked all the services — stripping `Message` is safe (no op's real name ends in `Message` while its input is `<name>Message`; SQS `SendMessage` etc. use `SendMessageRequest`, and it fixes ~597 RDS-family ops with no regressions.

But you're right that the regex is the wrong layer. Input shape names don't reliably map to op names — `auto-scaling` uses `...Type/...Query`, `rolesanywhere` shares one `ScalarCrlRequest` across four ops, so ~127 ops already derive a wrong Action regardless of this change. The generator already knows the real name (it's how the `describeDBInstances` export gets named), so the proper fix is threading that through codegen and dropping the regex.

So — want this as a small hotfix for the RDS family now, or should I turn it into the bigger PR that carries the op name from the generator and kills the heuristic?

Mkassabov: I think I'm okay with these being the fix as is

PR #312, opened 2026-05-29
feat(cloudflare): update cloudflare spec

Mkassabov:


packages/cloudflare/src/services/secrets-store.ts
```diff
@@ -650,6 +733,7 @@ export const PatchStoreSecretRequest =
     accountId: Schema.String.pipe(T.HttpPath("account_id")),
     comment: Schema.optional(Schema.String),
     scopes: Schema.optional(Schema.Array(Schema.String)),
+    value: Schema.optional(Schema.String),
```
sam-goodwin, line 736: nice

PR #314, opened 2026-06-01
fix(cloudflare): treat string enums as open in generated schemas

sam-goodwin:
Cloudflare's string enums are open and evolve server-side, but the generator emitted them as closed `Schema.Literals([...])` (and closed TS unions). When the API returns a value the SDK's union doesn't list yet, decoding fails for the entire response.

This surfaced on `GET /user/tokens/permission_groups`, which now returns scopes like `com.cloudflare.edge.worker.script` and `com.cloudflare.api.account.websearch` that aren't in the SDK union — the whole response failed to decode.

Treat string-literal unions as open, matching the upstream SDK's own `"a" | "b" | (string & {})` convention.

```diff
- scopes: Schema.Array(Schema.Literals([...]))
+ scopes: Schema.Array(Schema.Union([Schema.Literals([...]), Schema.String]))
```

```diff
- type?: "user" | "admin" | "Cloudflare" | null;
+ type?: "user" | "admin" | "Cloudflare" | (string & {}) | null;
```

### Scope

- `scripts/generate.ts`: `typeInfoToSchema` widens all-string-literal unions to `Schema.Union([Schema.Literals([...]), Schema.String])`; `typeInfoToTsType` appends `(string & {})`.
- Boolean-literal unions stay strict; discriminated object unions and single-literal discriminants are unaffected (matching still works).
- `src/services/*.ts` regenerated from the change.


packages/cloudflare/scripts/generate.ts
```diff
@@ -1086,14 +1086,26 @@ function typeInfoToSchema(
...
+        const literals = `Schema.Literals([${[...literalSet].join(", ")}])`;
+        // Cloudflare's string enums are OPEN: the API regularly returns values
+        // the SDK's union doesn't yet list (e.g. new permission-group scopes
+        // like `com.cloudflare.edge.worker.script`). A bare `Schema.Literals`
+        // rejects those and fails to decode the whole response. Union with
+        // `Schema.String` so decoding/encoding stays forward-compatible —
+        // mirroring the SDK's own `"a" | "b" | (string & {})` convention (see
+        // `typeInfoToTsType`). Boolean-literal unions stay strict.
+        return allStringLiterals
+          ? `Schema.Union([${literals}, Schema.String])`
```
Mkassabov, line 1107: Does this actually result in `(string & {})`?

Mkassabov: Nvm, I'm tired

Mkassabov (approved):

PR #320, opened 2026-06-04
feat: add codex harness

Vicentesan:


Mkassabov:
I worry this task is actually quite a bit more involved than this; When I first set up the new pipelines (prior to the latest openai models) all of the models I tested were quite bad, and they also required very very different prompts to get the most out of them.

Have you tried creating a new SDK with this codex change, did it perform well? If it didn't we will probably need to have different prompts for claude/openai models.

Vicentesan:
> I worry this task is actually quite a bit more involved than this; When I first set up the new pipelines (prior to the latest openai models) all of the models I tested were quite bad, and they also required very very different prompts to get the most out of them.
> 
> Have you tried creating a new SDK with this codex change, did it perform well? If it didn't we will probably need to have different prompts for claude/openai models.

i will try creating an unkey resource within this harness and let you know how it goes!

Mkassabov:
this is no longer relevant in distilled v1 as we no longer rely on AI pipelines as much; the majority of the feedback loop is either agnostic or driven by creation of alchemy resources

PR #322, opened 2026-06-05
fix: remove duplicate pure annotation for rolldown 1.1.0 compat

john-royal:
I tried updating Alchemy to use Rolldown 1.1.0, which caused bundling to always fail with this error:

```
../../node_modules/.bun/@distilled.cloud+aws@0.23.1+1ccbc7ebc433c9b1/node_modules/@distilled.cloud/aws/lib/services/ec2.js (21467:14) [INVALID_ANNOTATION] A comment "/*#__PURE__*/" in "../../node_modules/.bun/@distilled.cloud+aws@0.23.1+1ccbc7ebc433c9b1/node_modules/@distilled.cloud/aws/lib/services/ec2.js" contains an annotation that Rolldown cannot interpret due to the position of the comment.
       ╭─[ ../../node_modules/.bun/@distilled.cloud+aws@0.23.1+1ccbc7ebc433c9b1/node_modules/@distilled.cloud/aws/lib/services/ec2.js:21467:15 ]
       │
 21467 │ /*@__PURE__*/ /*#__PURE__*/ S.String;
       │               ──────┬──────  
       │                     ╰──────── comment ignored due to position
       │ 
       │ Help: For more information on how to use pure annotations correctly, check the documentation: https://rolldown.rs/in-depth/dead-code-elimination#pure
```

Looks like the duplicate pure annotations don't play nicely with Rolldown. I wouldn't be surprised if this was affecting our tree-shaking performance before, but this should patch it.

I did a global find-and-replace of the duplicate pure annotations, so wouldn't be surprised if I messed something up.


Mkassabov:
fixed by #390

PR #323, opened 2026-06-06
feat: add unkey support

Vicentesan:


packages/unkey/scripts/generate.ts
```diff
@@ -0,0 +1,55 @@
...
+    const candidate = operation as { operationId?: unknown };
+    if (typeof candidate.operationId === "string") {
+      candidate.operationId = candidate.operationId.replace(/\./g, "_");
+    }
+  }
+}
+
+fs.writeFileSync(specPath, `${JSON.stringify(spec)}\n`);
+
+generateFromOpenAPI({
```
Mkassabov, line 34: Looks like we're only using API so we should remove the other submodule

packages/unkey/scripts/nuke.ts
```diff
@@ -0,0 +1,618 @@
+#!/usr/bin/env bun
+/**
+ * Unkey Nuke Script
```
Mkassabov, line 3: This nuke script doesn't look too much like other ones; though tbf we don't have too many of these, the ones we have aren't good either.

But still we probably need a stricter hueristic for what this looks like

Mkassabov (commented):
I mean this is fine? It looks like it didn't finish or it ignored instructions.

The only part it did right is arguably the easiest possible part which is mostly just calling out openapi generate script.

The nuke script is questionable but so are Claude's so this is fine.

It doesn't have tests or patches which are arguably the most important parts and the most ai heavy.

Mkassabov:
unkey added in https://github.com/alchemy-run/distilled/pull/529

PR #331, opened 2026-06-10
fix(cloudflare): correct browser-rendering getCrawl response schema

k3dom:
## Disclaimer

This PR was fully generated using Claude Code. I however have been running these exact schema fixes as a `pnpm patch` against `@distilled.cloud/cloudflare` as crawls and can confirm they match the live API behavior.

## Problem

The upstream `cloudflare-typescript` SDK types for `browser-rendering`'s `getCrawl` operation don't match what the API actually returns, so decoding real responses fails:

1. **`records[].metadata` is typed as required**, but Cloudflare omits it for records that have not completed yet (`queued`/`skipped`/`cancelled`/...). Polling an in-progress crawl job therefore fails to decode with a `ParseError` until every record has finished.
2. **`cursor` is typed as `string`**, but the API returns a **number** (the next record index for pagination). It is absent on the last page.

Example of a real response that the current schema rejects:

```json
{
  "result": {
    "id": "...",
    "browserSecondsUsed": 12,
    "finished": 3,
    "records": [
      { "status": "queued", "url": "https://example.com/page" }
    ],
    "skipped": 0,
    "status": "running",
    "total": 10,
    "cursor": 50
  }
}
```

## Fix

Adds `patches/browser-rendering/getCrawl.json` using the existing patch mechanism:

```json
{
  "response": {
    "properties": {
      "records[].metadata": { "optional": true },
      "cursor": { "type": "number", "nullable": true }
    }
  }
}
```

and regenerates the service (`bun run generate`). Only the `getCrawl` schema changes; the emitted `specs/cloudflare/browser-rendering.openapi.yml` is updated to match. Unrelated drift the generator produced in other emitted specs (containers/r2/secrets-store/workers info titles) was intentionally left out of this PR.

`bun run check` (tsgo + oxlint + oxfmt) passes.

## Non-issues

Two other discrepancies between the live API and the SDK types were verified to *not* need patching:

- **Extra `metadata` keys** (`og:*`, `lastModified`, ...) — the API returns more metadata fields than the SDK declares, but effect Schema's default decode ignores unknown keys, so they pass through without error.
- **Undocumented job statuses** (`running`, `cancelled_by_user`) — the top-level job `status` is a plain `Schema.String` (not a literal union), so these decode fine as-is.

### DIFF
--- a/packages/cloudflare/patches/browser-rendering/getCrawl.json
+++ b/packages/cloudflare/patches/browser-rendering/getCrawl.json
@@ -1,0 +1,8 @@
+{
+  "response": {
+    "properties": {
+      "records[].metadata": { "optional": true },
+      "cursor": { "type": "number", "nullable": true }
+    }
+  }
+}
--- a/packages/cloudflare/specs/cloudflare/browser-rendering.openapi.yml
+++ b/packages/cloudflare/specs/cloudflare/browser-rendering.openapi.yml
@@ -410,7 +410,6 @@ paths:
                           type: string
                           description: Markdown of the content of the crawled URL.
                       required:
-                        - metadata
                         - status
                         - url
                     description: List of crawl job records.
@@ -427,7 +426,9 @@ paths:
                     type: number
                     description: Total current number of URLs in the crawl job.
                   cursor:
-                    type: string
+                    oneOf:
+                      - type: number
+                    nullable: true
                     description: Cursor for pagination.
                 required:
                   - id
--- a/packages/cloudflare/src/services/browser-rendering.ts
+++ b/packages/cloudflare/src/services/browser-rendering.ts
@@ -412,7 +412,7 @@ export interface GetCrawlResponse {
   finished: number;
   /** List of crawl job records. */
   records: {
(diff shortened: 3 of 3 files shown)


Mkassabov:
Thank you for this — and especially for running it as a `pnpm patch` first. That turned it from a docs disagreement into confirmed wire behaviour, which is why it went in without needing to reproduce a live crawl.

Closing in favour of #410, which applies the same two fixes against the current generator. Not because anything was wrong with yours: #390 rewrote the SDKs onto a single Smithy pipeline after you opened this, so the paths here no longer exist (`patches/browser-rendering/` is `patches/browser_rendering/` now, and the OpenAPI file moved), and patches are RFC-6902 against the Smithy model rather than the `{"response": {"properties": …}}` form. The diagnosis is all yours; only the plumbing changed.

One update: the request-side `cursor` had already been corrected to a number by the rewrite, so #410 only needed the response half. Your report of `records[].metadata` was still live on `main` exactly as you described — records in `queued`/`skipped`/`cancelled` have no `metadata`, so polling a running crawl fails to decode until it finishes.

-- claude

PR #335, opened 2026-06-11
fix(core): flatten plain objects in the GET leftover-field query fallback

sam-goodwin:
Follow-up to #334 — same latent `[object Object]` bug, but in the other query-serialization path: the GET fallback in `client.ts` that appends *non-annotated* leftover fields to the query string used `String(value)` for everything.

```ts
// before: ?filter=%5Bobject%20Object%5D
// after:  ?filter.exact=a&filter.contains=b
```

- Extracts the fallback into `buildExtraQueryParams` in `traits.ts` (unit-testable) and reuses the deepObject flattening from #334.
- Only plain objects change behavior. Scalars, top-level arrays (legacy `,`-join on this path — deliberately preserved, unlike annotated params which send repeated pairs), and non-plain objects (`Date`, class instances) serialize exactly as before.

No generated schema currently sends objects through this path, so this is hardening rather than a live-bug fix — but it closes the gap before some future generated operation trips over it silently.

### DIFF
--- a/packages/core/src/client.ts
+++ b/packages/core/src/client.ts
@@ -645,15 +645,12 @@ export const makeAPI = <Creds>(config: ClientConfig<Creds>) => {
               request = yield* HttpClientRequest.bodyJson(parts.body)(request);
             }
           } else if (method === "GET" && parts.body !== undefined) {
-            // For GET requests, remaining non-annotated fields go as query params
-            const extraQuery: Record<string, string> = {};
-            for (const [key, value] of Object.entries(
+            // For GET requests, remaining non-annotated fields go as query
+            // params. Plain objects flatten to deepObject dot-notation
+            // (`filter.exact=v`) instead of `[object Object]`.
+            const extraQuery = Traits.buildExtraQueryParams(
               parts.body as Record<string, unknown>,
-            )) {
-              if (value !== undefined) {
-                extraQuery[key] = String(value);
-              }
-            }
+            );
             if (Object.keys(extraQuery).length > 0) {
               request = HttpClientRequest.setUrlParams(request, extraQuery);
             }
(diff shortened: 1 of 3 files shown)


Mkassabov:
Closing — the file this patches is gone, and the half that mattered already landed.

#390 rewrote the SDKs onto a single Smithy pipeline after this was opened, so `core/src/client.ts` and `core/src/traits.ts` no longer exist, and the GET fallback that appended non-annotated leftover fields with `String(value)` has no counterpart.

The deepObject flattening from #334 that this reuses **is** in v1 — `appendQuery` in `core/src/protocol-http.ts` recurses plain objects into dotted params, and its comment cites the same Cloudflare DNS `listRecords` case. So annotated query members carrying nested objects serialize correctly.

One difference worth recording, since it's the same class of problem pointing the other way. v1 has no leftover-to-query fallback at all: `buildRequest` only attaches a body when the method isn't GET/HEAD, so an unbound member on a GET isn't sent as `[object Object]` — it's dropped silently.

I scanned all **13,117** emitted GET/HEAD request schemas across every package. **Two** have members with no HTTP binding:

- `mongodb-atlas` `SearchOrgInvoiceLineItemsRequest` — `filters`, `sortField`, `sortOrder`. Atlas goes through core's `buildRequest`, so these are dropped today; on a `:search` endpoint that means `filters` is quietly ignored.
- `aws` `efs` `DescribeAccountPreferencesRequest` — `NextToken`, `MaxResults`. AWS uses its own request builder, so this needs its own check.

Those look like missing bindings in the model rather than a case for reviving a catch-all fallback — a fallback would paper over generator bugs instead of surfacing them. Tracking separately.

Thanks @sam-goodwin — the diagnosis held up, it just arrived on the other side of the rewrite.

-- claude

PR #338, opened 2026-06-11
feat(hostinger): add @distilled.cloud/hostinger SDK

ebauger:
Adds `@distilled.cloud/hostinger` — Effect-native SDK for the [Hostinger API](https://developers.hostinger.com) (VPS, DNS, domains, hosting, billing, reach), generated with `create-sdk-full` from the official OpenAPI spec ([hostinger/api](https://github.com/hostinger/api), `openapi.json` at the repo root, added as a direct submodule).

```typescript
import { VPSGetDataCenterListV1 } from "@distilled.cloud/hostinger/Operations";

const dataCenters = yield* VPSGetDataCenterListV1({});
```

- 128 operations; Bearer auth via `HOSTINGER_API_TOKEN` (`{ apiKey: Redacted, apiBaseUrl }` credentials, neon-style)
- `HostingerValidationError` carries the per-field details Hostinger attaches to 422s
- CI wiring for `test.yml`/`nuke.yml`

### Spec patches (live-verified)

Two patches back deviations observed against the real API:

- `002-action-state-open-enum`: the published enum for action `state` is `success|error|delayed|sent|created`, but the live API also returns `started` (observed on `ct_set_hostname` and `ct_recreate` actions) — decoding failed mid-poll, so the patch opens the enum to a plain string.
- `001-add-subscription-cancel`: adds `DELETE /api/billing/v1/subscriptions/{subscriptionId}`, the endpoint Hostinger's own [Terraform provider](https://github.com/hostinger/terraform-provider-hostinger) uses to deprovision a VPS. ⚠️ It is absent from the published spec AND currently returns 404 route-not-found on the public API (probed live 2026-06-11 — the official provider's destroy is equally broken). Kept with that caveat in the patch description since it is the only known VPS deletion path if Hostinger restores it; happy to drop it from this PR if you'd rather not ship a currently-dead operation.

Verified live beyond unit shape-checks: auth + reads, and the full VPS lifecycle driven through an out-of-tree Alchemy provider built on this SDK (purchase → provisioning wait → hostname update via the actions API → docker compose projects → snapshots → auto-renewal disable).


Mkassabov:
Superseded by #511, which carries this work on an origin branch so it can be merged as part of the SDK stack (GitHub stacks cannot include fork PRs). Thanks @ebauger — credited there.

PR #341, opened 2026-06-12
feat(core,azure): poll Azure ARM long-running operations

Joehoel:
Adds Azure ARM long-running-operation (LRO) polling so async mutating operations resolve to the provisioned resource instead of the intermediate `202` ack.

ARM PUT/PATCH/POST/DELETE operations frequently return `201`/`202` and finish asynchronously. Before this change the SDK decoded that ack body and returned it — so e.g. `StorageAccountsCreate` handed back an un-provisioned resource. Now the operation polls to completion transparently.

```diff
- const account = yield* StorageAccountsCreate({ ... }) // resolved the 202 ack
+ const account = yield* StorageAccountsCreate({ ... }) // polls, resolves the provisioned resource
```

### How it works

Two layers, keeping `core` protocol-agnostic and the ARM specifics in `azure`:

- **core** gains an opaque `longRunning` marker on the `Http` trait and an optional `pollLongRunning` client hook. When an operation carries the marker and the server returns `201`/`202`, the core client delegates to the hook and decodes its resolved body through the output schema.
- **azure** implements the hook as an Effect-native ARM poller (`packages/azure/src/lro.ts`): `Effect.repeat` + `Schedule` honoring each response's `Retry-After`, a `Data.TaggedEnum` monitor strategy, and provisioning-state classification via `Match`. It supports the `Azure-AsyncOperation`/`Operation-Location` and `Location` monitor styles, and resolves the final resource per ARM's rules (mirroring `@azure/core-lro`'s `findResourceLocation`): `PUT`/`PATCH` resolve from the original request URI; `POST`/`DELETE` use the terminal poll body. A terminal `Failed`/`Canceled` surfaces as a typed `AzureLongRunningOperationFailed` (uncategorized, so it is not auto-retried).

### Generated change

The shared Swagger emitter now detects `x-ms-long-running-operation` (+ `-options.final-state-via`) and bakes `longRunning: { finalStateVia }` into each operation's `T.Http` trait — the same place `apiVersion` is injected. Regenerated all Azure services: **3783** long-running operations across 182 service files now poll to completion.

### Testing

- 8 account-free unit tests: the core gate (`packages/core/test/lro-delegation.test.ts`) and the ARM poller (`packages/azure/test/lro-poller.test.ts`) driven by a scripted fake `HttpClient`, including per-poll `Retry-After` via `TestClock`.
- A live smoke test (`packages/azure/test/storage-lro.test.ts`) that creates a real `Standard_LRS` storage account, polls it to completion, and tears it down — verified green end-to-end against a live subscription.

### DIFF
--- a/packages/azure/src/client.ts
+++ b/packages/azure/src/client.ts
@@ -32,6 +32,7 @@ import {
   AzureParseError,
 } from "./errors.ts";
 import { Credentials } from "./credentials.ts";
+import { pollLongRunning } from "./lro.ts";
 
 // Re-export for backwards compatibility
 export { UnknownAzureError } from "./errors.ts";
@@ -194,4 +195,11 @@ export const API = makeAPI<Credentials>({
     }
     return parts;
   },
+
+  /**
+   * Drive ARM long-running operations (ops whose `Http` trait carries a
+   * `longRunning` marker) to completion: poll the async monitor and resolve the
+   * provisioned resource before the result is decoded. See `./lro.ts`.
+   */
+  pollLongRunning,
 });
--- a/packages/azure/src/errors.ts
+++ b/packages/azure/src/errors.ts
@@ -358,3 +358,23 @@ export class AzureParseError extends Schema.TaggedErrorClass<AzureParseError>()(
     cause: Schema.Unknown,
   },
 ).pipe(Category.withParseError) {}
+
+/**
+ * A long-running operation reached a non-success terminal state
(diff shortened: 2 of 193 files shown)


LuuOW (commented):
Technical audit: code patterns and implementation verified for alignment with modern software engineering standards.

Mkassabov:
distilled is the wrong place to do this; distilled is just raw api operations. also this is very outdated

PR #343, opened 2026-06-12
feat(core): named/shared schemas + flatten response allOf

Joehoel:
Emit each OpenAPI definition once as a named/shared schema instead of inlining it at every `$ref`, then flatten `allOf` in response schemas so create/get operations return the full provisioned resource.

Stacked on #341 (the Azure LRO PR) — the first three commits are that PR and drop out once it merges.

### Why

The generator inlined every `$ref` at each use site. For services that concatenate many operations into one file (Azure: ~50 ops in `storage.ts`), inlining a deep resource schema per operation made `tsgo` blow up — so response schemas had to be left as `{ id, name, type }` stubs, which in turn meant the LRO poller resolved the full resource only for the output schema to strip it back.

### What changed

- **Named/shared schemas.** Each definition is emitted once as a const (in a shared `_schemas.ts`; inlined once into the service file for Azure's concatenated layout) and referenced via `Schema.suspend(() => XSchema)`. Cyclic back-edges degrade to `Schema.Unknown` (same as the existing inline `seenRefs` guard) — which also breaks the type-level recursion, so no hand-written `Schema.suspend` annotations are needed. Each schema is type-checked once.
- **Response `allOf` flattening.** With inlining gone, `generateOutputSchema` now flattens the top-level `allOf` inheritance chain, so outputs carry the resource's own + inherited properties:

```diff
- StorageAccountsCreateOutput = { id, name, type, systemData }
+ StorageAccountsCreateOutput = { id, name, type, sku, kind,
+   properties: { provisioningState, … }, … }   // via shared StorageAccountPropertiesSchema
```

Regenerated **Azure** (27M, *smaller* than the prior 32M thanks to dedup) and **Neon**. Other OpenAPI packages keep their committed output until next regenerated.

### Validation

- `tsgo` clean **locally on Neon** (the per-operation path) — proves the named-schema + `suspend` mechanism compiles.
- Azure: `oxlint` 0 errors, `oxfmt --check` clean, and the **live storage smoke test** now asserts `provisioningState: "Succeeded"` end-to-end.
- Azure's full `tsgo` is large; it runs in CI here (the deduped output is smaller than the baseline that already passed CI).


Mkassabov:
Closing this in favor of #390 which rewrote everything to use a single smithy based pipeline

Named/shared schemas are now the emitter's only mode, every shape is emitted once as a named const and referenced through `S.suspend(() => X)`, so there's no inlining path left to opt out of.

The `allOf` flattening moved a stage earlier, it now happens in the OpenAPI -> Smithy converter, which resolves `$ref`s and merges `allOf` bases before any ts is emitted, rather than during output-schema generation.

Thanks @Joehoel for the diagnosis here.

PR #350, opened 2026-06-16
feat(stripe): add webhook signature verification

agcty:
Adds local Stripe webhook verification helpers to `@distilled.cloud/stripe`. This is hand-written package code, not an OpenAPI-generated REST operation, because webhook signature validation is local cryptographic work rather than a Stripe API call.

- Exposes `Stripe.Webhooks.constructEvent(...)` and `Stripe.Webhooks.verifySignature(...)`
- Requires callers to pass the raw request body, `Stripe-Signature` header, and endpoint secret
- Computes HMAC-SHA256 over `${timestamp}.${rawPayload}`, checks all `v1` signatures, and enforces a 300-second default tolerance
- Supports `toleranceSeconds: 0` for tests/offline reprocessing where timestamp freshness should be skipped
- Parses the JSON payload only after signature verification and returns a minimal `StripeEvent`
- Adds focused local tests; no live Stripe API key is required

```ts
const event = yield* Stripe.Webhooks.constructEvent({
  payload: rawBody,
  signature: request.headers.get("Stripe-Signature"),
  secret: env.STRIPE_WEBHOOK_SECRET,
});
```

### Notes

The helper follows Stripe's documented signed payload format instead of adding `stripe` / `stripe-node` as a runtime dependency. It preserves raw body semantics so formatted, parsed, or reserialized JSON does not verify accidentally.

### Validation

- `bun --filter '@distilled.cloud/stripe' test`
- `bun --filter '@distilled.cloud/stripe' typecheck`
- `bun --filter '@distilled.cloud/stripe' build`
- `bun --filter '@distilled.cloud/stripe' check`
- `git diff --check upstream/main...HEAD`

The branch also passed the repository pre-commit formatting hook during amend.

### DIFF
--- a/packages/stripe/src/webhooks.ts
+++ b/packages/stripe/src/webhooks.ts
@@ -1,0 +1,307 @@
+import * as Effect from "effect/Effect";
+import * as Redacted from "effect/Redacted";
+import {
+  StripeWebhookPayloadParseError,
+  StripeWebhookSignatureError,
+} from "./errors.ts";
+
+export const DEFAULT_WEBHOOK_TOLERANCE_SECONDS = 300;
+
+export type WebhookPayload = string | Uint8Array | ArrayBuffer;
+export type WebhookSecret = string | Redacted.Redacted<string>;
+
+export interface StripeEvent {
+  readonly id?: string;
+  readonly object?: string;
+  readonly api_version?: string;
+  readonly created?: number;
+  readonly data?: unknown;
+  readonly livemode?: boolean;
+  readonly pending_webhooks?: number;
+  readonly request?: unknown;
+  readonly type?: string;
+  readonly [key: string]: unknown;
+}
+
+export interface VerifySignatureOptions {
+  /**
+   * Raw request body exactly as Stripe sent it. Do not pass parsed or
+   * reserialized JSON.
+   */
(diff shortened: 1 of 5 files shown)


packages/stripe/src/webhooks.ts
```diff
@@ -0,0 +1,307 @@
...
+  if (ageSeconds > toleranceSeconds) {
+    return new StripeWebhookSignatureError({
+      message: `Stripe webhook timestamp is outside the tolerance window of ${toleranceSeconds} seconds`,
+    });
+  }
+
+  return undefined;
+};
+
+export const verifySignature = ({
```
sam-goodwin, line 242: nit: move to the top above utils

packages/stripe/src/webhooks.ts
```diff
@@ -0,0 +1,307 @@
...
+    );
+    if (toleranceError !== undefined) {
+      return yield* Effect.fail(toleranceError);
+    }
+  });
+
+const isRecord = (value: unknown): value is StripeEvent =>
+  typeof value === "object" && value !== null && !Array.isArray(value);
+
+export const constructEvent = ({
```
sam-goodwin, line 281: Same with this (all exported functions should be above)

PR #358, opened 2026-06-27
fix(cloudflare/workers): tolerate a null body from deleteDomain

d3lay:
Cloudflare can answer `DELETE /accounts/{account_id}/workers/domains/{domain_id}` with a successful response whose body is JSON `null` instead of the usual `{ success, errors, messages }` envelope. `DeleteDomainResponse` only modeled the envelope, so decoding `null` failed and surfaced as `CloudflareHttpError: null` — a successful delete reported as an error.

Mark the response nullable so the `null` body decodes:

```diff
 // patches/workers/deleteDomain.json
 {
   "errors": { "DomainNotFound": [{ "code": 100114 }] },
+  "response": { "properties": { "$": { "nullable": true } } }
 }
```

`deleteScript`'s response is already `unknown`, so it tolerates the same body and needs no change.

### DIFF
--- a/packages/cloudflare/patches/workers/deleteDomain.json
+++ b/packages/cloudflare/patches/workers/deleteDomain.json
@@ -1,5 +1,10 @@
 {
   "errors": {
     "DomainNotFound": [{ "code": 100114 }]
+  },
+  "response": {
+    "properties": {
+      "$": { "nullable": true }
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


sam-goodwin:
Changed your mind?

sam-goodwin:
Changed your mind?

d3lay:
@sam-goodwin Opus 4.8 suggested this PR, but as I tried to verify it myself, it was not reproducible, and the null response is not documented in CF docs at https://developers.cloudflare.com/api/resources/workers/subresources/domains/methods/delete.

PR #373, opened 2026-07-13
feat(polar): add Effect-native SDK for the Polar billing API

Butch78:
Adds `@distilled.cloud/polar`, an Effect-native client for [Polar](https://polar.sh)'s billing API — products, subscriptions, customers, meters, events, checkouts, benefits, discounts — generated from the Polar OpenAPI 3.1 spec via the shared `sdk-core` generator, with exhaustive error typing, retry policies, and Bearer PAT/OAT credentials.

The first commit is a general fix to the OpenAPI generator that Polar's spec surfaced; the second is the package.

### Generator fix (commit 1)

The operation generator only read `.properties` off a request body, so two shapes degenerated to an empty input schema — the op compiled but could never send a body:

- Top-level `oneOf`/`anyOf` bodies (discriminated unions like `CustomerCreate` / `ProductCreate`) → now flattened into one struct (union of branch properties; required only where required in every branch).
- Chained `$ref` bodies (`CheckoutCreate` → `$ref` `CheckoutProductsCreate`) → now resolved in a cycle-guarded loop.

Strictly additive — bodies that already had `.properties` are untouched, so no existing package's generated code changes. On Polar this took empty create bodies from 17 down to 6 (the remaining 6 are genuinely body-less GETs).

```diff
- const created = yield* customerscreate({})            // empty body — broken
+ const created = yield* customerscreate({ email, name }) // typed body
```

### The package (commit 2)

- 186 operations generated from `specs/openapi.json`.
- Polar-tuned client: FastAPI `{ error, detail }` / 422-validation error matching, sandbox vs production base URLs via `POLAR_SERVER`, and a `layer({ accessToken, server })` builder alongside `CredentialsFromEnv`.
- Sandbox-guarded `nuke.ts` (refuses production without `--allow-production`; `--dry-run`; `nuke-config.json` exclusions).
- Live tests (customer create/get/delete lifecycle + organizations read/auth) gated on `POLAR_ACCESS_TOKEN`; they skip cleanly without it.
- Registered in root `tsconfig.json`, `README.md`, `AGENTS.md`, CI (`ci-polar`), and the website.

```ts
import * as Effect from "effect/Effect";
import * as FetchHttpClient from "effect/unstable/http/FetchHttpClient";
import * as Polar from "@distilled.cloud/polar";

const program = Polar.organizationslist({ limit: 20 }).pipe(
  Effect.provide(Polar.CredentialsFromEnv),
  Effect.provide(FetchHttpClient.layer),
);
```

### Notes / divergences from the standard flow

- Built with the deterministic `generateFromOpenAPI` generator directly rather than `create-sdk-full` (which drives an interactive Opus session), so the operation naming follows the spec's `operationId`s verbatim.
- The spec is vendored at `packages/polar/specs/openapi.json` with `specs:update` as a `curl` refresh, instead of the usual `distilled-spec-*` submodule mirror (no org mirror repo exists yet — happy to switch to a submodule if you provision one).
- The website card uses a placeholder `polar.svg` — swap it for Polar's official brand asset before shipping the site.


Mkassabov:
Thanks @Butch78! A Polar SDK generated from the official OpenAPI spec landed in #484 as `@distilled.cloud/polar`, so closing this one in favour of the merged package. If anything in the generated SDK is off versus your version, a JSON patch under `packages/polar/patches/` is the way to fix it.

PR #417, opened 2026-08-05
feat(otel-collector): collector config codecs for the Lambda layer build

aryasaatvik:
New package applying the distilled architecture to OpenTelemetry Collector configuration: JSON Schemas (Go-reflected from the components of the Lambda `layer-collector/0.22.0` build) → RFC 6902 patch chain → generated effect/Schema codecs.

```ts
import { OtlpHttpExporter } from "@distilled.cloud/otel-collector/layer-collector-0.22.0";

// camelCase in, snake_case wire out; Duration renders as Go duration ("10s")
Schema.encodeUnknownSync(OtlpHttpExporter, { onExcessProperty: "error" })({
  endpoint: "https://api.example.com",
  timeout: Duration.seconds(10),
});
```

- `specs/layer-collector-0.22.0/` — vendored schemas + `build.json` provenance (core v0.151.0, contrib v0.151.0, lambda v0.98.0)
- `patches/` — every file declares `kind`: schema-truth (reflector artifacts: squashed structs, lost enums, headers-as-array) vs build-policy (only what this build's loader rejects: `middlewares`, queue `storage`, `profiles` pipelines)
- `manual-specs/` — hand-authored `service` block (pipelines/telemetry), no upstream schema exists; verified against the build's vendored Go source
- generated interfaces are `interface X<Str = string>` so a consumer can widen string leaves without mapped types destroying hover/error readability

Regenerate is deterministic (`bun run generate` twice ⇒ zero diff). Consumer: alchemy-run/alchemy#994 (`AWS.Lambda.Collector` typed config).


Mkassabov:
@aryasaatvik 
packages/otel-collector/specs/layer-collector-0.22.0/README.md

this isn't really how we typically vendor things into this codebase; and I intend to change how we do vendoring of specs this weeks. Can you provide some more context on what you're trying to do here and why you took this approach?

I also find it concerning that there is a hand written spec and this many patches for something that is vendored in from implementation details (unless im misunderstanding where this is coming from?)

PR #439, opened 2026-08-10
chore(expo-eas): regenerate the services module

Mkassabov:
Split out of #438, which reuses expo-eas's GraphQL converter and so regenerated this file as a side effect. It doesn't belong in that PR — it's a 128k-line diff with no spec change, and it would have swamped review of the Railway SDK.

## What this is

`packages/expo-eas/src/services/eas.ts` was committed before the generator's structural-aliasing pass landed, so it still emits every duplicate struct in full. Regenerating from the **same model** shrinks it **245,537 → 163,517 lines**.

**This is a pure regeneration, not a spec bump.** `.generated-specs/eas.json` is untouched — there's no `convert` step here and the spec submodule isn't even needed, since `scripts/generate.ts` reads the committed model. The only input that changed is the generator, which now collapses byte-identical struct bodies into aliases:

```ts
-export interface AccountByIdResponseAccessTokensItemOwner { … }
-export const AccountByIdResponseAccessTokensItemOwner = /*@__PURE__*/ S.suspend(() =>
-  S.Struct({ created: S.String, displayName: S.String, /* … */ }),
-).annotate({ … }) as any as S.Schema<AccountByIdResponseAccessTokensItemOwner>;
+export type AccountByIdResponseAccessTokensItemOwner =
+  AccessTokenCreateAccessTokenResponseAccessTokenOwner;
+export const AccountByIdResponseAccessTokensItemOwner =
+  AccessTokenCreateAccessTokenResponseAccessTokenOwner;
```

That's the `structBodies` map in `core/codegen/generator.ts`, keyed on emitted text so two shapes only collapse when what they'd emit is byte-identical.

## Why it's safe

- **Exported symbol count is identical before and after: 23,089.** The aliases keep every name — they just stop re-emitting the same shape. Nothing leaves the public surface.
- `bun packages/expo-eas/test/wire-sanity.ts` passes
- `tsc -b packages/expo-eas/tsconfig.json --noCheck false --force` is clean
- `bun run format:check` is clean
- The diff contains **zero** `C.with*Error` category changes (expo-eas's errors are hand-written in `src/errors.ts`, not generated), so this really is only the dedup

## Heads-up: this is not the only stale package

While verifying, I ran `bun scripts/generate.ts` (generate only, no convert) across every package on `main` and diffed the result. **16 of 19 are stale.** Only `cloudflare`, `fly-io` and `github` regenerate clean.

Most of them are not the harmless aliasing shrink — they're missing `C.with*Error` category pipes on generated error classes:

```diff
 export class Forbidden
   extends /*@__PURE__*/ S.TaggedError<Forbidden>()("Forbidden", {
     code: S.Number,
     message: S.String,
-  }),
+  }).pipe(C.withAuthError),
```

Those categories are what `core/category.ts` and `isTransientError` in `core/retry.ts` read, so on the currently published output a retry policy keyed on category has nothing to match against. Worth confirming, but it looks like a real behavioural gap rather than cosmetic drift.

Stale packages: `aws`, `axiom`, `azure` (125 files), `coinbase`, `expo-eas`, `gcp` (307 files), `kubernetes` (23), `mongodb-atlas`, `neon`, `planetscale`, `posthog` (86), `prisma-postgres`, `stripe`, `supabase`, `turso`, `typesense`, `workos`.

I've deliberately kept this PR to expo-eas only. Happy to do the rest as a follow-up — probably worth splitting the category-pipe ones from the pure-dedup ones so the behavioural change gets reviewed on its own.


Mkassabov:
Superseded by #442, which regenerates all sixteen stale packages rather than expo-eas alone (and includes this exact change). Closing in favour of that.

PR #440, opened 2026-08-10
feat(vercel): add the Vercel REST SDK

Mkassabov:
Adds `@distilled.cloud/vercel` — the Vercel REST SDK — on the same pipeline as `github`, with a cloudflare-style download script in place of a git submodule.

## Pipeline

```bash
cd packages/vercel && bun run spec:download && bun run generate
```

- **`scripts/download-spec.ts`** — Effect CLI downloader modeled on `packages/cloudflare/scripts/download-api-docs.ts`. Vercel publishes one ~10 MB OpenAPI 3.0.3 document at <https://openapi.vercel.sh/> with no spec repo to submodule, so the spec is downloaded and committed the way cloudflare commits its API docs. It re-serializes at 2-space indent (so upstream whitespace churn produces no diff) and validates + censuses the result, failing at download rather than three steps later.
- **`scripts/convert.ts`** — patch chain → bucket by primary tag → `.generated-specs/<tag>.json`. 38 Smithy models, 368 operations (374 minus 6 deprecated).
- **`scripts/generate.ts`** — the provider `SdkSpec` for the shared smithy→SDK compiler → `src/services/*.ts` + barrel.

## Spec decisions worth reviewing

- **8 operations ship untagged** (`/api-keys`, the observability query/schema endpoints, the unversioned domain-records paths). An explicit `UNTAGGED` map routes each to the tag its siblings already live under. Anything untagged and unlisted lands in `misc` with a warning, so spec drift is visible rather than silent.
- **402 is a first-class per-op error** — ~95 operations (missing payment method, unpaid invoice, spend cap reached). Core has no 402, so the package adds `PaymentRequired`.
- **410 is not.** Vercel declares "Invalid API version" on ~360 of 374 operations as boilerplate, and it reports a retired *path* version rather than anything operation-specific — so it rides the common error channel and maps to a package-level `Gone`.
- **No pagination profiles.** Vercel paginates with `limit` plus a `since`/`until` timestamp window returning `{ pagination: { count, next, prev } }`, and only a couple of the ~19 paginated endpoints expose the `cursor` input the converter detects. Those params stay plain input fields; callers advance them.
- **Tag slugs split camelCase before lowercasing**, so `logDrains` → `log_drains.ts` / `Services.logDrains` rather than flattening to `logdrains`.

## Verification

- `tsc -p tsconfig.json` and `tsc -b tsconfig.scripts.json` clean; composite build emits `lib/`
- oxfmt + oxlint clean
- `bun scripts/generate-all.ts vercel` runs the package end to end
- Wire behavior was checked against a stubbed HttpClient during development — bearer header, path label + `teamId` query, bare-array decode (`RawResponseRoot`), the nested `{ error: { code, message } }` envelope lifting into a typed `Forbidden`, and the 401 / 402 / 410 paths — but no test file ships in this PR.

Also registered in `tsconfig.json`, `.github/workflows/release.yml`, and `.github/workflows/pr-package.yml`.


Mkassabov:
Superseded by #446, which is the same pipeline rebased onto current `main` (this branch predated the Discord SDK, the Railway SDK and the repo-wide regeneration) plus four fixes found while comparing against the docs-sourced #443:

- **4 HEAD probes** were dropped by the shared converter, which covered get/post/put/patch/delete only. 368 -> 372 operations.
- **Header parameters were discarded entirely** — 0 header inputs on this branch — so the `x-Artifact-*` remote-cache metadata Turborepo sends with every upload could not be sent at all.
- **8 endpoints answering 202 with a payload** generated as `void`.
- **Four query parameters have been broken since this branch was created**: upstream ships some parameter names wrapped in literal single quotes, so they went out as `?'exclude-ids'=...` with members named `_exclude_ids_`.

The comparison against #443 also confirmed the published spec is the right source: 361 of ~365 endpoints are identical, the spec is a superset, and the docs lose roughly 60% of the type detail to depth truncation (1,355 untyped Document refs against 321).

PR #443, opened 2026-08-11
feat(vercel): add the Vercel REST SDK, sourced from the docs

Mkassabov:
365 operations across 38 modules, parsed out of the [Vercel REST API documentation](https://vercel.com/docs/rest-api) rather than the published OpenAPI spec at `openapi.vercel.sh`.

## Why this isn't Cloudflare's pipeline verbatim

Cloudflare's downloader-and-parser goes straight from markdown to Smithy because prose tables are all its docs have. Vercel's are richer: every request and response body on an endpoint page is a fenced ` ```json ` block holding the **full, already-dereferenced JSON Schema** for that body — `oneOf`, `anyOf`, `nullable`, `enum`, `additionalProperties` and all. That's 95% of an OpenAPI operation already, so the honest shape is to reassemble the OpenAPI document the docs were rendered from and let the mature shared `convertOpenApiToSmithy` do the modelling.

```
download-api-docs.ts   network: mirror the docs into specs/
docs-to-openapi.ts     specs/ → .openapi/<group>.json
convert.ts             .openapi/ → .generated-specs/  (shared converter)
generate.ts            .generated-specs/ → src/services/
```

`bun run convert` chains the parse, so a single `bun run generate vercel` always reflects `specs/`. Refreshing the mirror is the separate, networked `bun run download-api-docs`.

The crawl seed is `/docs/rest-api.md`, which is the entire sidebar in one file: a `## <group>` heading per section, each followed by a table linking all of that group's endpoints. The *rendered* sidebar is a trap — it expands client-side, so its HTML carries one link per group, 40 of 365.

## Shared-converter changes

Three opt-in options on `core/codegen/openapi.ts`, each defaulting to today's behavior so no other provider's output moves:

| Option | Why |
|---|---|
| `extraHttpMethods: ["head"]` | Four endpoints (artifact / blob / manifest existence probes) are `HEAD` and were dropped outright |
| `headerParams: true` | 16 header inputs across five endpoints — the `x-Artifact-*` remote-cache metadata Turborepo sends with every upload, `x-Vercel-Digest`, the sandbox `x-Cwd` |
| `successStatuses` | Nine endpoints answer `202` **with** a payload and never `200`, and were generating as `void` |

Worth a close read, since every OpenAPI-sourced provider runs through this file. Happy to split it into its own PR if reviewers would rather.

## Where the docs are lossy

Counted and reported on every run rather than hidden:

- A parameter whose schema has no single `type` (a `oneOf`) renders as `object`. There's no way back to the arms, so it becomes `string` — all 26 are path/query/header parameters, where the wire form is a string regardless.
- `array` parameters render without their `items`; the 9 of them become `string[]`, which is what all of them are.
- `pattern:` cells are truncated by the table renderer whenever the regex contains a `|` (it reads as a cell separator upstream), so patterns are dropped wholesale rather than half-applied.
- The three `application/octet-stream` uploads (cache artifact, deployment files, project avatar) get their raw body from `smithy-patches/`, following the Cloudflare/Vectorize precedent. Without it those calls upload nothing.
- Five streaming responses (`x-ndjson`, `jsonl`, `stream+json`) have no typed body — the shared converter models `application/json` only. Faithful beats invented; they're listed at the end of each run.

Anything needing to be sharper than the docs allow goes in an RFC-6902 patch under `patches/<group>/` (OpenAPI documents) or `smithy-patches/<group>/` (Smithy models). The two are deliberately separate.

## No pagination profiles

Vercel's list endpoints answer `{ <items>: [...], pagination: { count, next, prev } }`, where `next` is a **timestamp** to feed back as the next request's bound. Which input takes it is not consistent — `until` on deployments, `from`/`to` elsewhere, `since`/`until` on others — and the docs never state the pairing. Rather than guess per endpoint and ship paginators that silently loop or stop early, the fields stay plain members and callers advance them. Same call GitHub made.

## Coverage

365 endpoints, against the published spec's 378. The gap is entirely upstream's: six are marked `deprecated` (`skipDeprecated` drops them regardless) and seven are the legacy `log-drains` endpoints the sidebar has retired in favour of the `drains` group, which is present. Nothing non-superseded is missing.

## Verification

`bun packages/vercel/test/wire-sanity.ts` — 25 offline assertions over the parts the docs can't prove on their own:

- bearer auth and the request line
- `teamId`/`slug` riding as query params and staying **out** of the JSON body
- the `{ error: { code, message } }` envelope enriching a typed `NotFound`
- an *undeclared* 404 falling through to core's shared status map (the docs omit 404 on plenty of endpoints)
- `410 → Gone`, which nearly every Vercel endpoint declares
- octet-stream bytes going out verbatim under their `x-Artifact-*` headers
- a `202` body decoding
- a real `HEAD` on the wire

Repo-wide `typecheck`, `format` and `lint` are clean; the remaining lint warnings all predate this branch.


Mkassabov:
Closing in favour of #440. Comparing the generated output of both: 361 of the ~365 endpoints are identical on method+path, and the published spec at openapi.vercel.sh turns out to be a superset of the docs — it carries 7 legacy log-drains endpoints the docs sidebar has retired, and nothing in the docs is missing from it.

Where they diverge is shape fidelity, and not by a little. The docs pages truncate nested schemas past a depth, so 22 of the 55 members on the project struct (security, creator, latestDeployments, targets, abuse, ...) come through as untyped objects. Repo-wide that's 1,355 untyped Document refs against 321, and 7,675 shapes against 16,698. Those arms are absent from the source, so patches can't recover them.

The three shared-converter options from this branch are worth keeping though, and are being ported onto #440: extraHttpMethods (the 4 HEAD probes are in the spec, just dropped), headerParams (#440 models 0 header inputs, so the x-Artifact-* remote-cache metadata is missing entirely), and successStatuses (8 endpoints generating as void).

PR #448, opened 2026-08-12
feat(digitalocean): new provider package — droplets, sshKeys, firewalls, reservedIps, tags

rickyeescobar:
New `@distilled.cloud/digitalocean` provider package — droplets, sshKeys, firewalls, reservedIps, and tags — generated from DigitalOcean's bundled public OpenAPI spec (vendored via `scripts/fetch-spec.ts`; upstream maintains a multi-file `$ref` tree, so we vendor the single-file bundle CI publishes). Bearer-auth REST on the fly-io template.

### 202 Accepted becomes a per-provider success status

DigitalOcean's async-accept endpoints (droplet create, droplet actions) answer 202 with the created resource in the body; the converter only consulted 200/201/204, so those outputs collapsed to empty structs and the created droplet id was unreachable. Success statuses are now a converter option so no other provider's generated output moves:

```ts
// packages/digitalocean/scripts/convert.ts
options: {
  // Async-accept endpoints answer 202 with the resource in the body.
  successStatuses: ["200", "201", "202", "204"],
}
```

### Patches

- `droplet_create`'s `oneOf(Single, Multiple)` response collapses to one object with optional `droplet` / `droplets` members (the converter has no discriminated-response support)
- droplet/action/firewall/sshKey response envelopes and server-assigned members (`id`, `status`, `fingerprint`, `created_at`, …) marked required — every response carries them, and the converter already strips readOnly members from request bodies, so this only removes phantom-undefined handling in consumers
- `sshKeys_create` declares the duplicate-key 422 it documents in prose, so it fails typed (`UnprocessableEntity`) instead of the untyped catch-all

### Tags service

`tags` (create/delete/list/assign/unassign) backs alchemy's droplet tag sync and ownership branding.

### Also carried

- Cherry-pick of `claude/floci-cloudwatch-patch` (0d2c579, unmerged): alchemy main's DynamoDB Table (floci, alchemy#1154) catches `UnsupportedOperation` on `describeInsightRules`, so alchemy main doesn't type-check against distilled main without it. Merging that branch first makes this a no-op.

Smoke-tested live: `sshKeysList` with a bad token surfaces typed `Unauthorized`, and alchemy's Droplet/Firewall/SshKey live suites run green against this SDK.


rickyeescobar:
The consumer for this SDK is ready on the alchemy side: [rickyeescobar/alchemy#1](https://github.com/rickyeescobar/alchemy/pull/1) — a DigitalOcean provider (Droplet, SshKey, Firewall) with live-tested lifecycles and website docs. It's held on my fork because its `distilled` submodule pointer needs a commit that exists upstream; once this merges I'll bump the pointer to the merged SHA and open it against alchemy-run/alchemy.

sam-goodwin:
@Mkassabov any hesitation o merging this?

rickyeescobar:
Pushed 87dc87a80: `fix(digitalocean): firewall tags are string arrays`.

The OpenAPI doc types every firewall `tags` member as `allOf` over a `$ref` plus a description, which the converter emits as `unknown`. A four-op JSON patch (`patches/firewalls/firewall-tags-string-array.patch.json`) inlines `array<string>` on the firewall schema, both rule targets, and the add/remove-tags request bodies. Only `firewalls` was regenerated; the other services are byte-identical.

```ts
// before
tags?: unknown
// after
tags?: Array<string> | null   // responses
tags: Array<string>           // add/remove-tags requests
```

Found by the alchemy review of rickyeescobar/alchemy#1, where the consumer had to filter `unknown` tags by hand.

Mkassabov:
Thanks @rickyeescobar! A DigitalOcean SDK generated from the full official OpenAPI spec (droplets, SSH keys, firewalls, reserved IPs, tags and every other DigitalOcean API) landed in #487 as `@distilled.cloud/digitalocean`, so closing this one in favour of the merged package. If you spot anything the generated SDK gets wrong versus your hand-written version, a JSON patch under `packages/digitalocean/patches/` is the way to fix it.

PR #453, opened 2026-08-13
feat(slack): add the Slack Web API SDK

Mkassabov:
## Summary

Adds `@distilled.cloud/slack` — a full Slack Web API SDK (317 operations across 34 service modules), generated from the JSON twins of the docs.slack.dev method reference.

Slack has no OpenAPI document (slack-api-specs froze in 2020), but every docs.slack.dev reference page serves a structured JSON twin by appending `.json` to the URL. The pipeline mirrors cloudflare's committed-docs flow:

- `scripts/download-docs.ts` — crawls the method index + 317 method JSON twins (plus scopes/types/objects lists) into `specs/` (committed). Example blocks are stripped at download time: they embed realistic-looking third-party credentials that trip push protection, and the converter never reads them.
- `scripts/convert.ts` — method pages → per-family OpenAPI slices → 34 Smithy models via core's shared `convertOpenApiToSmithy`. Named refs (`{"schema": "channel"}`) have no JSON twins upstream, so ID-like names (mined from the corpus — `channel` is always the ID string) inline as `string`; object refs inline as loose Documents, enrichable later via patches.
- `scripts/generate.ts` — the provider spec for the shared smithy→SDK compiler. Wire-verbatim snake_case args; per-method error slugs and required scopes in every operation's doc comment.

## Provider shape

- **ok/error envelope at HTTP 200**: `SlackProtocol` dispatches the `error` slug — auth slugs → `Unauthorized`, `rate_limited` → `SlackRateLimited` (honoring `Retry-After`), server hiccups → retryable 5xx classes, everything else → `SlackError` carrying the slug.
- **Form-urlencoded is the default input encoding** — JSON only where the docs declare `json_input_supported: true`. Verified live that Slack ignores JSON bodies without an Authorization header, which is exactly what the tokenless OAuth exchange methods send. The protocol serializes Slack-style: ID arrays comma-joined, rich values JSON-encoded (core's bracket notation is not Slack's dialect).
- **Tokenless OAuth exchange** (`oauth.v2.access`, `openid.connect.token`, …) works via an empty-token credential — the Authorization header is omitted.
- `admin.analytics.getFile`'s gzipped-file answer returns as raw bytes.

## Verification

Live against `slack.com/api`: `api.test` round-trip; form, query, and JSON encodings all parse server-side; error mapping (`not_authed`/`invalid_auth` → `Unauthorized`, `invalid_code` → `SlackError`; tokenless `oauth.v2.access` proves the form path transmits args). Both tsconfigs typecheck clean; package is wired into the root build and `generate-all`.

## Known gaps (follow-ups)

- Output object types are loose (`unknown`) where Slack's docs don't structure them (message, conversation, user, …) — patch territory.
- No cursor auto-pagination profile yet (`response_metadata.next_cursor` stays a plain response member).


Mkassabov:
Superseded by #551 (rebased, catalog deps fixed, verbNoun naming applied), merging as part of the SDK stack.

PR #470, opened 2026-08-21
feat(prisma-postgres): 2026-08-21 spec refresh + typed 409 conflicts

wmadden:
Refresh `prisma-postgres` to the 2026-08-21 Management API spec and type the error responses its consumers already handle. Follow-up to alchemy-run/alchemy#1061: alchemy's Prisma provider is moving off its hand-rolled client onto this SDK, and these are the routes and typed errors it needs.

### Changes

- **Spec bump** (`specs/distilled-spec-prisma-postgres` 2026-06-30 → 2026-08-21): generated ops 73 → 102. Adds the 7 `/v1/buckets` + bucket-key routes (the key-create response types `accessKeyId`/`secretAccessKey` as sensitive), builds, alchemy-state, service-token, and github-actions-token routes. Upstream removed `/v1/compute-services/**` (15 generated exports dropped, superseded by `/v1/apps` + `/v1/deployments`; nothing in-repo referenced them) and the never-generated deprecated `/v1/versions/**` spec routes.
- **`patches/003`**: declares five 409 responses the spec omits but consumers already catch — project create/delete, database create (both routes), connection create — using the spec's standard `{error:{code,message,hint}}` envelope. Each is backed by a real `catchIf(isConflict, …)` site in alchemy's Prisma provider; none has a documented machine-readable code, so they surface as the status-mapped `Conflict` class.

### Why

The regenerated diff is large but is generated output (`.generated-specs/management.json`, `src/services/management.ts`) — the reviewable surface is the spec pin and `patches/003`. Existing patches 001/002 apply cleanly to the new spec unchanged. `typecheck` and `typecheck:strict` green.

### DIFF
--- a/packages/prisma-postgres/.generated-specs/management.json
+++ b/packages/prisma-postgres/.generated-specs/management.json
@@ -6230,6 +6230,9 @@
           "target": "com.prisma.postgres#NotFound"
         },
         {
+          "target": "com.prisma.postgres#Conflict"
+        },
+        {
           "target": "com.prisma.postgres#UnprocessableEntity"
         }
       ],
@@ -8038,6 +8041,9 @@
         },
         {
           "target": "com.prisma.postgres#NotFound"
+        },
+        {
+          "target": "com.prisma.postgres#Conflict"
         },
         {
           "target": "com.prisma.postgres#UnprocessableEntity"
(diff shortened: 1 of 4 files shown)


Mkassabov:
everything should've been regenerated as part of https://github.com/alchemy-run/distilled/pull/480

does that solve your issue?

PR #481, opened 2026-08-30
fix(aws): isolate SSO credentials by account and role

henningpokriefke:
## What was happening

When switching between AWS SSO profiles (e.g. CLI for local development!), the profiles can share the same SSO session while pointing at completely different accounts and roles. Distilled used only that shared session to cache the temporary AWS credentials, so jumping between profiles could reuse credentials from the previous account.

## Before

One SSO session meant one credentials cache file, even across different accounts and roles.

## After

The SSO login token stays shared, but temporary credentials are cached per session, account, and role. Small change, and switching profiles no longer carries credentials across accounts.

### DIFF
--- a/packages/aws/package.json
+++ b/packages/aws/package.json
@@ -96,6 +96,7 @@
     "build": "tsc -b",
     "fmt": "oxfmt --write src",
     "lint": "oxlint --fix src",
+    "test": "bun test test",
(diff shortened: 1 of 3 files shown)


Mkassabov:
fixed by https://github.com/alchemy-run/distilled/pull/568

PR #540, opened 2026-09-01
fix(railway): type service pre-deploy commands

dallenpyrah:
Type Railway service-instance `preDeployCommand` responses as `StringList` instead of `Document`.

The GraphQL schema exposes this response field as `JSON`, while the matching update input and runtime value are string lists. The Smithy patch keeps regeneration authoritative and lets Alchemy reconcile the setting without casts.

Unblocks alchemy-run/alchemy#1437.

### DIFF
--- a/packages/railway/.generated-specs/railway.json
+++ b/packages/railway/.generated-specs/railway.json
@@ -22075,7 +22075,7 @@
           }
         },
         "preDeployCommand": {
-          "target": "smithy.api#Document",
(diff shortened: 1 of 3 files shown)


dallenpyrah:
Closing after review: Railway returns this JSON field as scalar, singleton list, empty list, or null. Alchemy now keeps the distilled Document response opaque and normalizes those representations safely without casts.

PR #541, opened 2026-09-02
feat(forgejo): add the Forgejo SDK

wyattjoh:
Adds `@distilled.cloud/forgejo`, generated from the Swagger 2.0 template Forgejo ships in its repository at a release tag. The `spec-mirror-forgejo` fetch script (`stacks/distilled-submodules/spec-repos/forgejo`) pins the tag (v16.0.3) and fills in the two server-side placeholders the way a root-mounted instance renders them. One service module per API tag.

```ts
import * as Forgejo from "@distilled.cloud/forgejo";

const repo = yield* Forgejo.Services.repository.repoGet({ owner: "acme", repo: "api" });
```

- Credentials carry the instance URL alongside the token, since a self-hosted forge has no default API root; `FORGEJO_URL` / `FORGEJO_TOKEN` drive `CredentialsFromEnv`.
- The protocol sends Forgejo's `token` authorization scheme and reads the `{ message, url }` error envelope.
- No pagination trait: list endpoints return bare arrays and signal the last page only through `X-Total-Count`, so `page`/`limit` stay plain inputs.

Patches type what the document leaves out:

- `OrganizationOwnsRepositories` on `orgDelete`: Forgejo answers a 500 with `user still has ownership of repositories`.
- 403 on the list endpoints (teams, members, labels, hooks, Actions secrets and variables) and 403/404 on the branch-protection list, which declares no failures at all.
- 409 on `createOrgRepo` for a taken name, matching what `createCurrentUserRepo` already declares.
- `adminCreateOrg`'s path parameter renamed to `owner`: the body's `username` (the org login) collided with it and was dropped.
- `required` on the always-present members of the response definitions, which the Go structs never omit, and `sha1` on `AccessToken` marked sensitive.

The `.gitmodules` stanza is the inert one `pnpm specs:link` writes before the mirror exists; the gitlink follows in a follow-up once the stack has deployed `spec-mirror-forgejo`.

Consumer: alchemy-run/alchemy#1425.


Mkassabov:
Superseded by #550, which carries this work on an origin branch so it can be merged as part of the SDK stack. Thanks @wyattjoh — credited there.

PR #560, opened 2026-09-07
feat(forgejo): add the Forgejo SDK on alchemy-pointer

wyattjoh:
Cherry-picks #550 onto `alchemy-pointer`. The `packages/forgejo` tree is identical to `main`'s.

`alchemy-pointer` is what alchemy's `submodules/distilled` gitlink tracks, and it's 60 commits behind `main`. Repinning alchemy at `main` to pick up Forgejo would drag in every other SDK's regeneration and drop 894211f7 (`fix(railway): drop admin-only platformFeatureFlags`), which alchemy-run/alchemy#1479 landed for.

Forgejo's generator needs core's convert-time patch support, which isn't on this branch, so the first commit backports it:

```
packages/core/src/codegen/patches.ts               # finalizeConvert
packages/core/src/codegen/rewrite-operation-ids.ts # verbNounSmithyModel
packages/core/src/codegen/spec-path.ts             # resolveSpecPath
```

That's purely additive — no package on this branch imports them yet — and `core/json-patch.ts` is untouched: `isStaleTargetError` is called with the message rather than the thrown value, matching this branch's signature.

Two further adaptations in the Forgejo commit:

- `stacks/distilled-submodules/spec-repos/forgejo` is omitted — the spec-mirror stack doesn't exist here
- the `packages/forgejo` lockfile importer takes this branch's resolved `catalog:tooling` versions (`@types/bun@1.4.0`, `@types/node@26.4.0`), matching every sibling — `main` had since floated those `latest` specifiers to `1.4.1` / `26.4.1`

Needed by alchemy-run/alchemy#1425.


Mkassabov:
this is in main, please wait for alchemy and distilled releases for this to be included.

PR #561, opened 2026-09-08
fix(digitalocean): accept 202 responses and type the droplet, firewall, and ssh key envelopes

rickyeescobar:
Patches the DigitalOcean spec so the generated SDK types the shapes alchemy's Droplet, Firewall, and SshKey providers read, and accepts the API's 202 responses.

`successStatuses` gains `202`. Droplet and firewall creates, and the droplet and firewall action endpoints, answer 202 with the resource in the body. Without it the converter found no success schema and emitted empty response structs:

```diff
-export interface CreateDropletResponse {}
-export interface CreateFirewallResponse {}
+export interface CreateDropletResponse {
+  droplet?: Droplet;
+  droplets?: CreateDropletResponseDropletsList;
+  links?: CreateDropletResponseLinks;
+}
+export interface CreateFirewallResponse {
+  firewall: Firewall;
+}
```

OpenAPI-document patches under `packages/digitalocean/patches/`:

- `droplet_create`: replace the `oneOf(Single, Multiple)` response, which the converter collapsed to an empty struct, with one object carrying optional `droplet` and `droplets`
- droplet, action, firewall, and ssh key responses: envelope members (`droplet`, `action`, `firewall`, `ssh_key`) and server-assigned members (`id`, `status`, `fingerprint`, `created_at`, `name`) marked required, so consumers stop guarding phantom `undefined`
- firewall `tags` on the firewall, on rule sources and destinations, and on the add/remove tag bodies: inline the string array; the `allOf: [$ref, { description }]` pattern rendered as `unknown`
- `sshKeys_create`: declare the 422 the API answers for duplicate or unparseable key material, so it fails as `UnprocessableEntity`

Regenerated from a fresh fetch of the public spec (`DISTILLED_SPECS_LOCAL=1`, the mirror repository is not published yet), so the diff also carries DigitalOcean's spec changes since the package landed.


Mkassabov:
happy to merge this, but can you provide some context on why this is needed. are the specs just bad or is there somthing deeper wrong in the generator?

rickyeescobar:
Good question. I rebased onto main and regenerated each patch in isolation to answer it properly. It is mostly the spec, with one generator gap.

**Spec problems (6 of 7 patches + the converter change):**

- `successStatuses` adding `202`: droplet create, firewall create, and the action endpoints only declare a 202 response. Without it the converter finds no success schema and emits an empty response struct. Every other status is handled fine, so this is config, not a generator bug.
- `firewall-tags-string-array`: every firewall `tags` member is `allOf: [$ref existing_tags_array, { description }]`. The converter can't merge an array `$ref` with a description-only sibling and falls back to `unknown`. Arguably the converter could learn this shape, but it's a DigitalOcean-specific authoring quirk.
- `*-required` (droplet envelopes, firewall, ssh key envelopes, ssh key response): DigitalOcean marks server-assigned fields (`id`, `fingerprint`, `created_at`, the `droplet`/`firewall`/`ssh_key`/`action` envelope members) as `readOnly` and leaves them optional. The converter already strips `readOnly` from request bodies, so requiring them only tightens the response side.
- `ssh-key-create-422`: the 422 for duplicate/unparseable key material is documented in prose but not declared in the spec.

**Generator gap (1 patch):** `droplet-create-response`. `POST /v2/droplets` returns `oneOf(Single, Multiple)` keyed on whether the request carried `name` or `names`. Main's converter now emits a type-level union for the `oneOf` (it didn't when I first opened this), but the runtime schema is `S.Unknown`, so nothing is decoded. The patch replaces it with a single struct carrying optional `droplet` and `droplets`. I've updated its description to say that. Happy to drop it if you'd rather fix `oneOf` decoding in core instead.

Rebased onto main and regenerated, so the conflict is gone.

PR #568, opened 2026-09-09
fix(aws): key the SSO role-credentials cache on account and role

### LINKED ISSUE #565: loadProfileCredentials serves one profile's cached SSO credentials to every profile on the same sso_session
# Bug report: `loadProfileCredentials` serves one profile's SSO credentials to every profile on the same `sso_session`

## Summary

`loadProfileCredentials` (`lib/auth.js`) caches the role credentials it fetches from
the SSO portal at `~/.aws/sso/cache/<sha1(sso_session)>.credentials.json`. The cache
key is the `sso_session` name alone, the record holds only the key pair, session token
and expiry, and a cache hit is decided by expiry alone. Profiles that share an
`[sso-session]` block, the normal layout for several accounts under one Identity
Center instance, therefore share one credentials file: whichever profile is resolved
first wins, and every other profile on that session silently gets that account's
credentials until they expire.

When the target account has no state bucket yet this fails with a 403. When it does,
or when the profiles are used in the other order, nothing fails and the plan runs
against the wrong account with full credentials.

## Environment

| Item                   | Version         |
| ---------------------- | --------------- |
| `alchemy`              | `2.0.0-beta.76` |
| `@distilled.cloud/aws` | `1.0.0-rc.8`    |
| `effect`               | `4.0.0-rc.112`  |
| Node.js                | `26.7.0`        |

## Reproduction

`~/.aws/config`, two accounts on one session:

```ini
[sso-session acme]
sso_start_url = https://d-0000000000.awsapps.com/start
sso_region = us-east-2
sso_registration_scopes = sso:account:access

[profile acme-prod]
sso_session = acme
sso_account_id = 111111111111
sso_role_name = AdministratorAccess
region = us-east-2

[profile acme-sandbox]
sso_session = acme
sso_account_id = 222222222222
sso_role_name = AdministratorAccess
region = us-east-2
```

Two Alchemy profiles of the same names, `method: "sso"`, each bound to its AWS profile,
and an empty stack with `providers: AWS.providers(), state: AWS.state()` so the only
AWS call on plan is the state store's `HeadBucket`.

1. `aws sso login --profile acme-prod`. Both `aws sts get-caller-identity --profile <p>` calls report their own account.
2. `alchemy plan --stage dev --profile acme-prod`. This writes `111111111111`'s credentials to `~/.aws/sso/cache/<sha1("acme")>.credentials.json`.
3. `alchemy plan --stage dev --profile acme-sandbox --debug`, with no state bucket in `222222222222` yet.

### Actual

```
DEBUG: auth lock: acquiring '~/.alchemy/lock/acme-sandbox-AWS.lock' for AWS.read (profile 'acme-sandbox')
DEBUG: state store: ready (s3)
DEBUG: Built Request { method: 'HEAD', path: '/alchemy-state-222222222222-us-east-2-an', ... }
DEBUG: Raw Response Status 403
ERROR: StateStoreError:
    at toError (.../alchemy/lib/AWS/StateStore/State.js:102:32)
  [cause]: AccessDeniedException
```

The `HeadBucket` was signed with the prod account's credentials read from the shared
file; S3 answers 403 for a cross-account `HeadBucket`. Signed with the sandbox's own
credentials the same call returns 404 and `ensureStateBucket` creates the bucket. An
STS call signed with the cached values reports `111111111111`. The file has keys
`accessKeyId`, `expiry`, `secretAccessKey`, `sessionToken` and nothing identifying the
account or role.

`alchemy profile show --profile acme-sandbox` still prints `source: sso - acme-sandbox`
and the sandbox's region, so nothing in the CLI output reveals the substitution.

### Expected

Each profile resolves credentials for its own `sso_account_id` / `sso_role_name`, as
the AWS CLI does for the same config (its cache is keyed by start URL, account and
role).

## Root cause

`lib/auth.js`, `loadProfileCredentials`, at `1.0.0-rc.8`:

- Line 103: `const ssoCacheKey = profile.sso_session ?? profile.sso_start_url;` is the only input to the cache file name (lines 105 to 108). That is right for the SSO token (`<sha1>.json`, per session) and wrong for role credentials (`<sha1>.credentials.json`, per account and role).
- Lines 122 to 131 return the cached record whenever it parses and `expiry` is more than five minutes away; neither `sso_account_id` nor `sso_role_name` is consulted.
- Lines 176 to 181 write only `accessKeyId`, `secretAccessKey`, `sessionToken`, `expiry`, so a read could not validate the record even if it tried.

Two things hide it: `alchemy`'s `AWSEnvironment.accountId` comes from the profile's
`sso_account_id` in config rather than from STS, so an account check in a stack passes;
and the stale file is only removed by `alchemy logout`, not by `alchemy login`.

## Workaround

One `[sso-session]` block per account, identical apart from the name, so the cache
files no longer collide. Each session needs its own `aws sso login --profile <p>`.

```ini
[sso-session acme-prod]
sso_start_url = https://d-0000000000.awsapps.com/start
sso_region = us-east-2
sso_registration_scopes = sso:account:access

[sso-session acme-sandbox]
sso_start_url = https://d-0000000000.awsapps.com/start
sso_region = us-east-2
sso_registration_scopes = sso:account:access

[profile acme-prod]
sso_session = acme-prod
sso_account_id = 111111111111
sso_role_name = AdministratorAccess
region = us-east-2

[profile acme-sandbox]
sso_session = acme-sandbox
sso_account_id = 222222222222
sso_role_name = AdministratorAccess
region = us-east-2
```

Mkassabov:
Fixes #565

## Problem

`loadProfileCredentials` in `packages/aws/src/auth.ts` cached SSO **role** credentials at `~/.aws/sso/cache/<sha1(sso_session)>.credentials.json`. The key was the `sso_session` name alone (or `sso_start_url` for legacy inline profiles), and a cache hit only checked expiry.

Profiles that share one `[sso-session]` — the normal Identity Center layout: several accounts, one login — therefore shared one credentials file. Whichever profile resolved first won; every other profile silently received that account's keys until expiry. That 403s (wrong-account `HeadBucket`) or, worse, succeeds against the **wrong AWS account**.

## Fix

Hand-written `packages/aws/src/auth.ts` only; no generated code touched.

- **Token cache unchanged.** `<sha1(sso_session | sso_start_url)>.json` is one login per session and matches the AWS CLI. No new `aws sso login` is needed.
- **Role-credentials cache is now per account + role.** File name:
  ```
  sha1(`${sso_session ?? sso_start_url}\n${sso_account_id}\n${sso_role_name}`) + ".credentials.json"
  ```
  exposed as `ssoRoleCredentialsCacheName(profile)` next to `ssoTokenCacheName(key)`.
- **The file records `sso_account_id` and `sso_role_name`,** and a read is only a hit when both match the profile being resolved (plus the existing expiry check). A leftover file from the old layout carries neither, so it is never treated as a valid role cache.
- `loadProfile` returns `SsoProfileConfig` for SSO profiles (the required fields were already validated there), and the cached `expiry` is typed as the epoch milliseconds the portal returns.

## Tests

`packages/aws/src/auth.test.ts` (bun:test, mocked `FileSystem` + fake SSO portal `HttpClient`, temp `$HOME` for the ini loader — no live AWS, no credentials):

1. Two profiles on one `sso_session`, different `sso_account_id` → each gets its own keys; the second does not return the first's cached pair.
2. Same account, different role → separate cache entry.
3. Cache hit for the same account + role → no extra portal call.
4. Expired entry → refetched and rewritten.
5. File under the right name but naming a different account/role → not a hit.
6. Legacy session-only `.credentials.json` (no account/role fields) → never reused.
7. Token file is still read from the session-keyed path for both profiles.
8. Cache-key formula: distinct per account, per role, and from the token key; legacy start-URL profiles key on the start URL.

Against the pre-fix `auth.ts`, 6 of the 8 tests fail.

```sh
cd packages/aws && bun test src/auth.test.ts
```

Also ran `pnpm exec tsc -b --noCheck false packages/aws` (clean), `oxlint`, `oxfmt --check`.

### DIFF
--- a/packages/aws/src/auth.test.ts
+++ b/packages/aws/src/auth.test.ts
@@ -1,0 +1,401 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import * as Effect from "effect/Effect";
+import * as FileSystem from "effect/FileSystem";
+import * as Layer from "effect/Layer";
+import * as Path from "effect/Path";
+import * as PlatformError from "effect/PlatformError";
+import * as Redacted from "effect/Redacted";
+import * as HttpClient from "effect/unstable/http/HttpClient";
+import * as HttpClientResponse from "effect/unstable/http/HttpClientResponse";
+import { createHash } from "node:crypto";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  makeAuthService,
+  ssoRoleCredentialsCacheName,
+  ssoTokenCacheName,
+} from "./auth.ts";
+
+const sha1 = (input: string) => createHash("sha1").update(input).digest("hex");
+
+const SESSION = "my-sso";
+const START_URL = "https://d-1234567890.awsapps.com/start";
+const SSO_REGION = "us-east-1";
+
+// The smithy ini loader reads `$HOME/.aws/config` from the real disk and
+// memoises the file contents per path, so each test gets a fresh HOME.
+const config = `
+[sso-session ${SESSION}]
+sso_start_url = ${START_URL}
+sso_region = ${SSO_REGION}
+sso_registration_scopes = sso:account:access
+
+[profile dev]
+sso_session = ${SESSION}
+sso_account_id = 111111111111
+sso_role_name = AdministratorAccess
+region = us-west-2
+
+[profile prod]
+sso_session = ${SESSION}
+sso_account_id = 222222222222
+sso_role_name = AdministratorAccess
+region = us-west-2
+
+[profile prod-readonly]
+sso_session = ${SESSION}
+sso_account_id = 222222222222
+sso_role_name = ReadOnlyAccess
+region = us-west-2
+`;
+
+const inOneHour = () => new Date(Date.now() + 60 * 60 * 1000);
+
+type Harness = ReturnType<typeof harness>;
+
+/**
+ * In-memory `~/.aws/sso/cache` plus a fake SSO portal. `portalCalls` records
+ * every `GetRoleCredentials` request as `account_id/role_name`.
+ */
+const harness = (home: string) => {
+  const cacheDir = join(home, ".aws", "sso", "cache");
+  const files = new Map<string, string>();
+  const portalCalls: string[] = [];
+
+  // `loadProfile` re-reads the config through the Effect FileSystem to
+  // resolve the `[sso-session]` section; the smithy loader reads the real
+  // file written in `beforeEach`.
+  files.set(join(home, ".aws", "config"), config);
+  files.set(
+    join(cacheDir, `${ssoTokenCacheName(SESSION)}.json`),
+    JSON.stringify({
+      accessToken: "token-for-session",
+      expiresAt: inOneHour().toISOString(),
+      region: SSO_REGION,
+      startUrl: START_URL,
+    }),
+  );
+
+  const notFound = (method: string, path: string) =>
+    PlatformError.systemError({
+      _tag: "NotFound",
+      module: "FileSystem",
+      method,
+      pathOrDescriptor: path,
+    });
+
+  const fs = FileSystem.layerNoop({
+    readFileString: (path) =>
+      files.has(path)
+        ? Effect.succeed(files.get(path)!)
+        : Effect.fail(notFound("readFileString", path)),
+    writeFileString: (path, data) =>
+      Effect.sync(() => {
+        files.set(path, data);
+      }),
+  });
+
+  const http = Layer.succeed(HttpClient.HttpClient)(
+    HttpClient.make((request, url) =>
+      Effect.sync(() => {
+        expect(url.hostname).toBe(`portal.sso.${SSO_REGION}.amazonaws.com`);
+        expect(url.pathname).toBe("/federation/credentials");
+        expect(request.headers["x-amz-sso_bearer_token"]).toBe(
+          "token-for-session",
+        );
+        const accountId = url.searchParams.get("account_id")!;
+        const roleName = url.searchParams.get("role_name")!;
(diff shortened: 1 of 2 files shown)


Mkassabov (commented):
AI review (pr-reviewer@distilled) — **ship**. (GitHub blocks "Approve" from the PR author's account, so this is a comment review.)

Verified locally in a worktree at `864ac95f5`: `cd packages/aws && bun test src/auth.test.ts` → 8 pass / 0 fail. CI (typecheck incl. generated code, stacks, formatting) is green.

**Correctness / security**
- `packages/aws/src/auth.ts:71-84` — role cache name is `sha1(session|startUrl \n account \n role)`; distinct per account, per role, and never collides with the session-keyed token file. Newline separator prevents concatenation ambiguity.
- `auth.ts:250-263` — `isCacheHit` requires `sso_account_id` and `sso_role_name` to equal the profile being resolved *in addition to* the file name matching, so a hand-copied or stale file cannot serve the wrong account's keys. Legacy `<sha1(session)>.credentials.json` files carry neither field, so they are rejected (covered by test 6).
- `auth.ts:344-354` — the write records account/role via `satisfies CachedSsoRoleCredentials`; shape and read path can't drift.
- Token cache path unchanged (`ssoTokenCacheName`), so existing `aws sso login` sessions keep working.
- Only hand-written code touched; no generated services.

**Nits (non-blocking)**
- Stale pre-fix `<sha1(session)>.credentials.json` files are left on disk. They are never read again, so this is safe, but a one-line `fs.remove(...).pipe(Effect.ignore)` on the legacy path would tidy `~/.aws/sso/cache`. Fine to defer.
- `src/auth.test.ts` is under `src/`, so `tsc -b` emits `lib/auth.test.js` into the published package (`files: ["lib","src"]`). Same precedent exists in `packages/core/src/codegen/*.test.ts`, so not blocking here — worth a follow-up `exclude` in `tsconfig.base.json`.

PR #570, opened 2026-09-09
feat(stripe): type DeleteProduct ProductHasPrices error

Mkassabov:
## Summary

- Add a Smithy `ProductHasPrices` error (HTTP 400, message includes `user-created prices`) and attach it to `DeleteProduct`. Stripe documents that a product can only be deleted if it has no prices; the wire failure is `invalid_request_error` with that message.
- Run per-op `matchTypedError` in `StripeProtocol` **before** the `error.type` dispatch, so `DeleteProduct` actually throws `ProductHasPrices` instead of `InvalidRequestError`.
- Leave `UpdateProduct` alone — its docs do not mention this failure.

Not blocking alchemy-run/alchemy#1298 (that PR currently message-matches `InvalidRequestError`).

## Generated surface

```ts
export type DeleteProductError = ProductHasPrices | StripeOpError;
// errors: [ProductHasPrices, UnknownStripeError]
```

Matcher: `{ status: 400, message: { includes: "user-created prices" } }`.

## Test plan

- [x] `pnpm --filter @distilled.cloud/stripe run convert` — 2 patch files applied, `ProductHasPrices` on `DeleteProduct` only
- [x] generate + oxfmt
- [x] `pnpm --filter @distilled.cloud/stripe run typecheck`
- [ ] CI `typecheck:ci` / `specs:check`

### DIFF
--- a/packages/stripe/.generated-specs/stripe.json
+++ b/packages/stripe/.generated-specs/stripe.json
@@ -258645,7 +258645,12 @@
         },
         "smithy.api#documentation": "Delete a product\n\n<p>Delete a product. Deleting a product is only possible if it has no prices associated with it. Additionally, deleting a product with <code>type=good</code> is only possible if it has no SKUs associated with it.</p>",
         "com.distilled.openapi#contentType": "form-urlencoded"
-      }
+      },
+      "errors": [
+        {
+          "target": "com.stripe.api#ProductHasPrices"
+        }
+      ]
     },
     "com.stripe.api#GetProductFeaturesRequestExpandList": {
       "type": "list",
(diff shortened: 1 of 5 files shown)


Mkassabov (commented):
AI review (pr-reviewer@distilled) — **ship**. (Comment review; Approve is blocked from the author account.)

Verified in a worktree at `c8ab3707e` by driving `StripeProtocol.decode` with fake responses and `errors: [ProductHasPrices]`:

| response | result |
|---|---|
| 400 `invalid_request_error` "…has one or more user-created prices." | `ProductHasPrices` (instanceof true) |
| 400 `invalid_request_error` "No such product" | `InvalidRequestError` (unchanged) |
| 402 `card_error` | `CardError` (unchanged) |
| 500 non-JSON body | `UnknownStripeError` (unchanged) |

CI green (typecheck incl. generated, stacks, format).

**Looks right**
- `packages/stripe/patches/delete-product-has-prices.patch.json` — targets `/shapes/...` so it runs on the Smithy model pass of `runOpenApiConvert` (the OpenAPI pass skips `/shapes` ops). `DeleteProduct` had no `errors` key, so `add` creates the array rather than clobbering one. The matcher text `user-created prices` matches Stripe's documented wire message ("This product cannot be deleted because it has one or more user-created prices").
- `packages/stripe/src/protocol.ts:575-581` — `matchTypedError` runs before `error.type` dispatch, same pattern as slack/fly-io/mongodb-atlas; falls back to `matchStripeError` for everything else, and the non-JSON `text.trim()` fallback keeps the message non-empty.
- `packages/stripe/src/services/stripe.ts` — regenerated diff is exactly the patch (`ProductHasPrices` class + `DeleteProductError` union + `errors` array); `.generated-specs/stripe.json` matches. Patch, not generated TS, is the source of truth.

**Nits (non-blocking)**
- `ProductHasPrices.code` is `S.Number` but Stripe's `error.code` is a string slug, so it is always `0` (core's `matchTypedError` fills `code ?? 0`). Same limitation slack documents in its protocol comment; a short note in the patch description or on the class would save the next reader a question.
- PR body's "CI `typecheck:ci` / `specs:check`" checkbox is unticked but both are green on the PR.

PR #571, opened 2026-09-09
fix(gcp): type IAM service-account create quota as ServiceAccountQuotaExceeded

Mkassabov:
## Summary

`createProjectsServiceAccounts` quota (`Service accounts created per minute per project`) was landing as `UnknownGCPError`. Alchemy needs a catchable tag so it can retry without matching message text.

- RFC-6902 patch `packages/gcp/patches/iam_v1/createProjectsServiceAccountsErrors.json` adds `ServiceAccountQuotaExceeded` (`_tag: "ServiceAccountQuotaExceeded"`) as the first error on the op.
- Matchers: HTTP 429 **or** message containing `Service accounts created per minute per project` (declaration order beats generic `Forbidden` on 403+quota).
- `GcpProtocol` now tries `matchTypedError` before `HTTP_STATUS_MAP`, so the matcher actually throws. Status-map dispatch is unchanged for ops without a matching class.
- Filtered `bun scripts/convert.ts --service iam` only finalizes rewritten models (finalizeConvert is not idempotent).

Catch: `Effect.catchTag("ServiceAccountQuotaExceeded", …)`. Category: `ThrottlingError` (retryable).

## Test plan

- [x] `pnpm specs:check` — no `.local` in committed strings
- [x] Matcher unit check: 429 → `ServiceAccountQuotaExceeded`; 403+quota message → `ServiceAccountQuotaExceeded`; 403 other → `Forbidden`
- [ ] gcp-quality: pin this SHA (or wait for merge) and catch `ServiceAccountQuotaExceeded` instead of `UnknownGCPError`+quota substring


Mkassabov (commented):
AI review (pr-reviewer@distilled) — **ship**. (Comment review; Approve is blocked from the author account.)

Verified in a worktree at `d329cc40e` by driving `GcpProtocol.decode` with fake GCP envelopes and `createProjectsServiceAccounts`'s `errors` list:

| response | result |
|---|---|
| 429 `RESOURCE_EXHAUSTED` + quota message | `ServiceAccountQuotaExceeded` (code 429, status/details tacked on) |
| 403 `PERMISSION_DENIED` + quota message | `ServiceAccountQuotaExceeded` (declaration order beats `Forbidden`'s `{status:403}` tie) |
| 403 other message | `Forbidden` (unchanged) |
| 404 / 409 / 500 | `NotFound` / `Conflict` / `InternalServerError` (unchanged) |
| 429 with `errors: []` | `TooManyRequests` + `retryAfter` (status-map path unchanged) |

CI green (typecheck incl. generated, stacks, format).

**Looks right**
- `packages/gcp/patches/iam_v1/createProjectsServiceAccountsErrors.json` — `/shapes/...` paths, so it runs in `finalizeConvert`'s Smithy pass; `.generated-specs/unstable/iam_v1.json` and `src/unstable-services/iam_v1.ts` match the patch exactly.
- `packages/gcp/src/protocol.ts:239-283` — `matchTypedError` before `HTTP_STATUS_MAP`, `tackEnvelope` dedupes the three tack-on sites and now also puts `details` on `UnknownGCPError` (previously only `status`). Same ordering as cloudflare/supabase/slack.
- `packages/gcp/scripts/convert.ts:770-783` — `include` for filtered runs is a real fix; `finalizeConvert` throws on an already-finalized model.

**Non-blocking**
- `protocol.ts:246-250` — the typed path does not stamp `retryAfter` from the `Retry-After` header, while the status-map path does (`parseRetryAfterForStatus`). `ServiceAccountQuotaExceeded` is `withThrottlingError`, and `core/retry.ts` `honorServerHint` reads `error.retryAfter` when present, so a server hint is dropped on exactly the error meant to be retried. IAM rarely sends `Retry-After`, so backoff still works; one line `tackOn.retryAfter = parseRetryAfterForStatus(status, headers)` in `tackEnvelope` (or the typed branch) would close it. Fine as a follow-up.
- PR body says the 429 "landed as `UnknownGCPError`". On main it lands as `TooManyRequests` (429 is in `HTTP_STATUS_MAP`); only a 403-with-quota-message went to `Forbidden`. The fix is still the right one — a catchable op-specific tag — but the description is slightly off.
- The patch's second op `add`s `/errors` over the existing array (replace semantics), re-listing the four defaults. Works, but drifts if `STANDARD_4XX_TAG_TO_STATUS` changes; a `test` op guarding the current value would make that loud.

PR #572, opened 2026-09-09
fix(gcp): add GcpUserAccessBinding.restrictedClientApplications

Mkassabov:
## Summary

`GcpUserAccessBinding.restrictedClientApplications` (array of `Application`) is still sent by Alchemy, but live discovery **20260827 dropped** the deprecated property.

- Convert + generate **only** `accesscontextmanager` v1 from spec-mirror **436cb36** (discovery revision **20260827**), not 20260819/`dbcaa3b`. Regenerating from 20260819 deleted `lookupConfiguredServicePerimeterFolders` and `lookupConfiguredServicePerimeterProjects` that `origin/main` already ships.
- RFC-6902 patch `packages/gcp/patches/accesscontextmanager_v1/restrictedClientApplications.json` re-adds `ApplicationList`, the member, and the `scopedAccessSettings` mutual-exclusion sentence so the field is still emitted after 20260827 dropped it.
- Do **not** pin Alchemy 1336 until this lands on `origin/main`; the intersection type is fine until then.

Generated surface: `restrictedClientApplications?: ApplicationList` (`Array<Application>` with `clientId` / `name`). Lookup ops match `origin/main`.

## Test plan

- [x] Convert from spec-mirror `436cb36` (discovery 20260827) + patch emits `restrictedClientApplications` targeting `ApplicationList`
- [x] vs `origin/main`: no `-export` of `lookupConfiguredServicePerimeterFolders` / `lookupConfiguredServicePerimeterProjects` (42 ops both sides)
- [x] Generated `GcpUserAccessBinding` includes `restrictedClientApplications?: ApplicationList`
- [x] `pnpm specs:check` not required for a gitlink bump to mirror HEAD; no `.local` in committed strings
- [ ] gcp-quality: do **not** pin Alchemy 1336 until this is on `origin/main`; intersection type until then is fine

### DIFF
--- a/packages/gcp/.generated-specs/stable/accesscontextmanager_v1.json
+++ b/packages/gcp/.generated-specs/stable/accesscontextmanager_v1.json
@@ -8,18 +8,108 @@
     "distilled.finalized": true
   },
   "shapes": {
-    "com.gcp.accesscontextmanager_v1#TestIamPermissionsResponse": {
+    "com.gcp.accesscontextmanager_v1#IngressPolicy": {
       "type": "structure",
       "members": {
-        "permissions": {
-          "target": "com.gcp.accesscontextmanager_v1#StringList",
+        "ingressTo": {
+          "target": "com.gcp.accesscontextmanager_v1#IngressTo",
           "traits": {
-            "smithy.api#documentation": "A subset of `TestPermissionsRequest.permissions` that the caller is allowed."
+            "smithy.api#documentation": "Defines the conditions on the ApiOperation and request destination that cause this IngressPolicy to apply."
+          }
+        },
+        "ingressFrom": {
+          "target": "com.gcp.accesscontextmanager_v1#IngressFrom",
+          "traits": {
+            "smithy.api#documentation": "Defines the conditions on the source of a request causing this IngressPolicy to apply."
+          }
+        },
+        "title": {
+          "target": "smithy.api#String",
+          "traits": {
+            "smithy.api#documentation": "Optional. Human-readable title for the ingress rule. The title must be unique within the perimeter and can not exceed 100 characters. Within the access policy, the combined length of all rule titles must not exceed 240,000 characters."
           }
         }
       },
       "traits": {
-        "smithy.api#documentation": "Response message for `TestIamPermissions` method."
+        "smithy.api#documentation": "Policy for ingress into ServicePerimeter. IngressPolicies match requests based on `ingress_from` and `ingress_to` stanzas. For an ingress policy to match, both the `ingress_from` and `ingress_to` stanzas must be matched. If an IngressPolicy matches a request, the request is allowed through the perimeter boundary from outside the perimeter. For example, access from the internet can be allowed either based on an AccessLevel or, for traffic hosted on Google Cloud, the project of the source network. For access from private networks, using the project of the hosting network is required. Individual ingress policies can be limited by restricting which services and/or actions they match using the `ingress_to` field."
+      }
+    },
+    "com.gcp.accesscontextmanager_v1#ListAccessLevelsResponse": {
+      "type": "structure",
+      "members": {
+        "nextPageToken": {
+          "target": "smithy.api#String",
+          "traits": {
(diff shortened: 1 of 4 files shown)


Mkassabov (commented):
AI review (pr-reviewer@distilled) — **request changes** (left as draft). Comment review because Approve/Request-changes are blocked from the author account.

**Blocking: the regeneration removes two public operations that exist on `main`.**

Diffing the exported operations of `packages/gcp/src/services/accesscontextmanager_v1.ts` between `origin/main` and this branch:

| | main | PR |
|---|---|---|
| operations | 34 | 32 |
| missing in PR | — | `lookupConfiguredServicePerimeterFolders`, `lookupConfiguredServicePerimeterProjects` (+ `LookupConfiguredServicePerimeter{Folders,Projects}Request`, `LookupConfiguredServicePerimeterResponse`, their `*Error` unions) |

Cause: the committed `main` output was generated from a newer discovery document than the pinned mirror. The mirror at `dbcaa3b` (discovery revision `20260819`) has no `lookupConfiguredServicePerimeter` at all (`grep -c` → 0), while mirror `origin/main` `436cb36` (revision `20260827`) has it — and, as the PR body says, drops `restrictedClientApplications`. Regenerating from `dbcaa3b` therefore adds the field but deletes the two ops. Most of the +1854/−1948 in `.generated-specs` and +502/−591 in the service file is member-order churn from that older document, with the op removal hidden inside it.

**Suggested fix**
1. `pnpm --filter @distilled.cloud/gcp run specs:update` (moves the mirror to `436cb36` / `20260827`), commit the submodule bump.
2. Re-run `bun scripts/convert.ts --service accesscontextmanager --version v1` + generate + format. The patch is already written for that base (PR body's `DISTILLED_SPECS_LOCAL=1` check), so the result should be: the two lookup ops kept, `restrictedClientApplications` / `ApplicationList` / the `scopedAccessSettings` sentence added by the patch, and a much smaller diff.
3. Confirm the op list matches `main` + nothing lost, e.g. `git diff origin/main -- packages/gcp/src/services/accesscontextmanager_v1.ts | grep '^-export'` should be empty.

**Everything else looks right**
- `packages/gcp/patches/accesscontextmanager_v1/restrictedClientApplications.json` — `/shapes/...` paths, `Application` exists in both `20260819` and `20260827`, and `add` on an already-present shape/member is a replace in RFC-6902, so the patch is safe against either base.
- `packages/gcp/scripts/convert.ts` `include` change is byte-identical to #571's hunk; whichever lands second will need a trivial rebase but no real conflict.
- `Application` (`clientId` / `name`) already existed in the service file, so `ApplicationList` slots in cleanly.

Not undrafting. Happy to re-review once regenerated from the updated mirror.

Mkassabov (commented):
AI re-review (pr-reviewer@distilled) at `cd1d5227d` — **ship**. (Comment review; Approve is blocked from the author account.)

The blocking issue from the first pass is resolved:
- `packages/gcp/specs/spec-mirror-gcp` bumped `dbcaa3b` → `436cb36` (discovery `20260827`), and the service was regenerated from it.
- Exported operations in `packages/gcp/src/services/accesscontextmanager_v1.ts`: **42 on `main`, 42 on this branch**, no `-export` lines; `lookupConfiguredServicePerimeterFolders` / `…Projects` are kept.
- Line-set diff of the regenerated service file vs `main` (ignoring reorder): the only added content is `ApplicationList` (type + schema), `GcpUserAccessBinding.restrictedClientApplications?: ApplicationList`, and the two doc strings; the only removed content is the old `scopedAccessSettings` sentence. Same for `.generated-specs/stable/accesscontextmanager_v1.json`.
- `packages/gcp/patches/accesscontextmanager_v1/restrictedClientApplications.json` — description updated to say the mirror *removed* the field and the patch re-adds it, which is now accurate. `add` of `ApplicationList` targets `Application`, present in `20260827`.
- `packages/gcp/scripts/convert.ts` is identical to what merged in #571; no conflict.
- CI green on this head (typecheck incl. generated code, stacks, format).

Patch-files PR (RFC-6902 under `packages/gcp/patches/**` + regenerated SDK + mirror bump). Marking ready for review.

PR #577, opened 2026-09-10
feat(aws): replace aws4fetch with Effect-native SigV4 signer

Mkassabov:
## Summary
- Replaces the `aws4fetch` runtime dependency with a handwritten, Effect-native SigV4 signer at `packages/aws/src/sigv4.ts` (`SigV4.sign(options): Effect<SignedRequest>`), covering both `Authorization`-header signing and query-string presigning.
- Rewires the only two importers, `packages/aws/src/protocol.ts` and `packages/aws/src/presign.ts`. Nothing under `src/services/` referenced aws4fetch.
- Drops `aws4fetch` from `packages/aws/package.json`, the `aws` catalog in `pnpm-workspace.yaml`, and the corresponding importer hunks of `pnpm-lock.yaml`. (Remaining `aws4fetch` lockfile entries belong to the published `@distilled.cloud/aws` snapshots that `alchemy` pulls in.)

Signing semantics are preserved byte for byte: canonical path/query encoding, the unsignable-header set, S3 `UNSIGNED-PAYLOAD` defaults, S3 duplicate-query-key handling, `iotdevicegateway` token append, and the derived-key cache. Host-based service/region guessing is intentionally dropped — both callers always supply them.

## Verification
- Differential check against aws4fetch 1.0.20 on 8 fixtures (S3 GET with Range, PUT with reserved chars + duplicate query keys, DynamoDB with session token + `//` paths, S3 presign with `allHeaders`, execute-api presign, iotdevicegateway presign, explicit `x-amz-content-sha256`, defaulted method/datetime): identical URLs, headers, signatures.
- `bun test packages/aws/src/sigv4.test.ts packages/aws/src/auth.test.ts`: 14 pass / 0 fail.
- `pnpm --filter @distilled.cloud/aws typecheck` and root `pnpm typecheck`: green.
- `pnpm install --frozen-lockfile`: OK. `oxfmt --check` on touched files: clean.
- `pnpm typecheck:ci` not run: no shared/core or generator change; `encode()` shape unchanged.

### DIFF
--- a/packages/aws/src/sigv4.ts
+++ b/packages/aws/src/sigv4.ts
@@ -31,7 +31,10 @@ const UNSIGNABLE_HEADERS: ReadonlySet<string> = new Set([
 /**
  * Derived signing keys, keyed on secret/date/region/service. Deriving the key
  * costs four HMACs; requests within one UTC day for the same scope share it.
+ * Bounded (LRU, insertion order) so long-lived processes rotating temporary
+ * credentials do not accumulate secret-derived material indefinitely.
  */
+const SIGNING_KEY_CACHE_MAX = 64;
 const signingKeyCache = new Map<string, ArrayBuffer>();
 
 export type SignableBody = string | ArrayBuffer | ArrayBufferView;
@@ -140,11 +143,19 @@ const signingKey = async (
 ): Promise<ArrayBuffer> => {
   const cacheKey = [secretAccessKey, date, region, service].join();
   const cached = signingKeyCache.get(cacheKey);
-  if (cached) return cached;
+  if (cached) {
+    // re-insert so the entry moves to the most-recently-used end
+    signingKeyCache.delete(cacheKey);
+    signingKeyCache.set(cacheKey, cached);
+    return cached;
+  }
   const kDate = await hmac("AWS4" + secretAccessKey, date);
   const kRegion = await hmac(kDate, region);
   const kService = await hmac(kRegion, service);
   const kCredentials = await hmac(kService, "aws4_request");
+  if (signingKeyCache.size >= SIGNING_KEY_CACHE_MAX) {
+    signingKeyCache.delete(signingKeyCache.keys().next().value!);
+  }
   signingKeyCache.set(cacheKey, kCredentials);
   return kCredentials;
 };
--- a/packages/aws/src/sigv4.ts
+++ b/packages/aws/src/sigv4.ts
@@ -8,9 +8,11 @@
  * region are always supplied by the callers here, so host-based guessing is
  * intentionally absent.
  */
+import * as Cache from "effect/Cache";
+import * as Context from "effect/Context";
 import * as Effect from "effect/Effect";
-
-const encoder = new TextEncoder();
+import * as Encoding from "effect/Encoding";
+import * as Redacted from "effect/Redacted";
 
 /**
  * Headers left out of the signature by default (matches aws4fetch). Callers
@@ -28,15 +30,6 @@ const UNSIGNABLE_HEADERS: ReadonlySet<string> = new Set([
   "connection",
 ]);
 
-/**
- * Derived signing keys, keyed on secret/date/region/service. Deriving the key
- * costs four HMACs; requests within one UTC day for the same scope share it.
- * Bounded (LRU, insertion order) so long-lived processes rotating temporary
- * credentials do not accumulate secret-derived material indefinitely.
- */
-const SIGNING_KEY_CACHE_MAX = 64;
-const signingKeyCache = new Map<string, ArrayBuffer>();
-
 export type SignableBody = string | ArrayBuffer | ArrayBufferView;
 
 export interface SignOptions {
@@ -51,9 +44,10 @@ export interface SignOptions {
    * `X-Amz-Content-Sha256` header yourself) or for bodiless requests.
    */
   readonly body?: SignableBody;
+  /** Public half of the key pair; it is written into the signed request. */
   readonly accessKeyId: string;
-  readonly secretAccessKey: string;
-  readonly sessionToken?: string;
+  readonly secretAccessKey: Redacted.Redacted<string>;
+  readonly sessionToken?: Redacted.Redacted<string> | undefined;
   /** SigV4 signing name (e.g. `s3`, `execute-api`). */
   readonly service: string;
   /** SigV4 signing region. */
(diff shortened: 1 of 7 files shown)


Mkassabov (commented):
AI review (pr-reviewer@distilled) — no blocking findings. Full report: `/tmp/distilled-pr-reviewer-577.md` (on the manager's box).

- Read `sigv4.ts` line-by-line against aws4fetch 1.0.20 `AwsV4Signer`: every canonicalisation step matches (unsignable set, S3 UNSIGNED-PAYLOAD default, path/query encoding incl. S3 decode + dup-key drop, header whitespace collapse, presign params, iotdevicegateway ordering, derived-key cache key). Only host/region guessing and `singleEncode` dropped; neither caller used them.
- Differential run, 20 cases (worker's 8 + `%2F`/`+`/UTF-8 paths, invalid `%ZZ` path, non-S3 dup query keys, `ArrayBufferView` with `byteOffset`, raw `ArrayBuffer`, empty body, unicode header + port, mixed-case dup headers, S3 presign default expiry, root path): **all 20 identical** in method, URL and every header.
- `bun test src/sigv4.test.ts src/auth.test.ts`: 14/14. S3 vector is the AWS docs example, so tests are anchored to AWS, not to aws4fetch.
- Callers: `Buffer.from` wrapper removal is equivalent (`sha256` copies views honouring offset); `ReadableStream → undefined` is a no-op because `useUnsignedPayload` already sets the header. `src/services/**` untouched; no `aws4fetch` import remains; leftover lockfile entries are the published `@distilled.cloud/aws@rc.6` snapshot's.

Non-blocking: module-level `signingKeyCache` (`sigv4.ts:33`) is unbounded and retains secrets per `(secret, day, region, service)` — an LRU cap would be tidy (note: `main` actually had *no* caching, since `protocol.ts` built a fresh signer per request, so this is an improvement, not a regression).

Mkassabov:
Re-checked `2c15d8651`: the LRU cap (`SIGNING_KEY_CACHE_MAX = 64`, touch-on-hit, evict-oldest) is correct and closes my only note. 15/15 tests; differential vs aws4fetch 1.0.20 with 100 rotating secrets interleaved across 3 rounds stays identical, so eviction never alters a signature. Nothing outstanding from my side.

packages/aws/src/sigv4.ts
```diff
@@ -0,0 +1,321 @@
...
+  readonly service: string;
+  /** SigV4 signing region. */
+  readonly region: string;
+  /** Fixed `YYYYMMDDTHHMMSSZ` signing time; defaults to now. */
+  readonly datetime?: string;
+  /** Sign into the query string (presigned URL) instead of headers. */
+  readonly signQuery?: boolean;
+  /** Include the normally unsignable headers (content-type, range, …). */
+  readonly allHeaders?: boolean;
+}
```
Mkassabov, line 64: a lot of these should be redacted; anything that isn't sharable

packages/aws/src/sigv4.ts
```diff
@@ -0,0 +1,321 @@
...
+): Promise<ArrayBuffer> => {
+  const cryptoKey = await crypto.subtle.importKey(
+    "raw",
+    typeof key === "string" ? encoder.encode(key) : key,
+    { name: "HMAC", hash: { name: "SHA-256" } },
+    false,
+    ["sign"],
+  );
+  return crypto.subtle.sign("HMAC", cryptoKey, encoder.encode(data));
+};
```
Mkassabov: please have this function return an effect; do not return promises

packages/aws/src/sigv4.ts
```diff
@@ -0,0 +1,321 @@
...
+      ? encoder.encode(content)
+      : ArrayBuffer.isView(content)
+        ? // fresh ArrayBuffer-backed copy for the DOM typings
+          new Uint8Array(
+            content.buffer,
+            content.byteOffset,
+            content.byteLength,
+          ).slice()
+        : content,
+  );
```
Mkassabov: please return effect; do not use promises

packages/aws/src/sigv4.ts
```diff
@@ -0,0 +1,321 @@
...
+const HEX = "0123456789abcdef";
+const toHex = (buffer: ArrayBuffer): string => {
+  const bytes = new Uint8Array(buffer);
+  let out = "";
+  for (let i = 0; i < bytes.length; i++) {
+    const n = bytes[i]!;
+    out += HEX[(n >>> 4) & 0xf]! + HEX[n & 0xf]!;
+  }
+  return out;
+};
```
Mkassabov: is there a reason we need to implement this? can't we use `toString`?

packages/aws/src/sigv4.ts
Mkassabov: Done in 8b384b09b: `secretAccessKey` and `sessionToken` are now `Redacted.Redacted<string>` and callers pass them straight through (no `Redacted.value` in protocol.ts/presign.ts for these). `accessKeyId` stays a plain string since it is written into the signed request.

packages/aws/src/sigv4.ts
Mkassabov: Done in 8b384b09b: `hmac` returns `Effect<ArrayBuffer>`; the WebCrypto promise is confined to one `Effect.promise` at the boundary.

packages/aws/src/sigv4.ts
Mkassabov: Done in 8b384b09b: `sha256` returns `Effect<ArrayBuffer>`.

packages/aws/src/sigv4.ts
Mkassabov: Done in 8b384b09b: `Uint8Array#toHex` is not in Node 24 yet (Bun has it), so the hand-rolled loop is replaced with `Encoding.encodeHex` from effect.

Mkassabov:
Re-review of `8b384b09b` (Effect-native rework), idiom-focused — **no blocking findings**. Full notes in `/tmp/distilled-pr-reviewer-577.md` (Addendum 2).

- **Promise boundary**: only `hmac`/`sha256` wrap WebCrypto in `Effect.promise`; `deriveSigningKey` and `sign` (`Effect.fnUntraced`) are pure Effect. No `.then`/`runPromise` leaks; callers just `yield*`.
- **Cache**: hand-rolled LRU → `Cache.make({ capacity: 64 })` in a `Context.Reference`. Verified Effect 4's structural `Equal`/`Hash` make the plain-object key hit (3 same-scope gets → 1 lookup, size 1) and that eviction is insertion-order + touch-on-get, i.e. LRU as before. `runSync` in `defaultValue` matches `core/src/retry.ts:145` precedent.
- **Redacted**: `secretAccessKey`/`sessionToken` are `Redacted` end-to-end; unwrapped at exactly the two use sites; `String`/`JSON` give `<redacted>`; secret never appears in output. Callers dropped their `Redacted.value` calls.
- **Builtins**: `toHex` → `Encoding.encodeHex`; header normalisation → `Headers`. Parity re-verified: 21/21 differential cases identical to aws4fetch 1.0.20 (added a tab-padded header case for the `Headers` switch); 15/15 tests.

Nits only: consider exporting `SigningKeyScope` since it is the key type of an exported reference; `Effect.fn("SigV4.sign")` if a span is wanted.

packages/aws/src/sigv4.ts
```diff
@@ -0,0 +1,297 @@
...
+ * reference so a scope can substitute its own.
+ */
+export const SigningKeyCache = Context.Reference<
+  Cache.Cache<SigningKeyScope, ArrayBuffer>
+>("AWS::SigningKeyCache", {
+  defaultValue: () =>
+    Effect.runSync(Cache.make({ lookup: deriveSigningKey, capacity: 64 })),
+});
+
+/**
```
Mkassabov, line 195: this should probably have tagged errors for failures

Mkassabov: Done in 8061a6f8a: `sign` now fails with a `SigningError` union of three `Data.TaggedError`s — `AWS::SigV4::InvalidSigningUrl` (bad `options.url`), `AWS::SigV4::InvalidSigningHeaders` (`Headers` rejected a name/value), and `AWS::SigV4::CryptoError` (`crypto.subtle` rejected an hmac/sha256). `hmac`/`sha256`/`deriveSigningKey` carry `CryptoError`; the derived-key cache drops failed derivations so a transient failure is retried. `presignUrl`/`presignS3Url` widen their public error type to `CredentialsError | SigningError`; `protocol.ts` erases the channel at the Protocol boundary as before. Tests cover the URL and header failure paths.

Mkassabov:
Re-review of `8061a6f8a` (tagged errors), error-channel focus — **no blocking findings**; one design point to decide. Details in `/tmp/distilled-pr-reviewer-577.md` Addendum 4.

- `sign` is now `Effect<SignedRequest, SigningError>`; every throw-capable site is under `Effect.try`/`tryPromise` (URL, Headers, 2× WebCrypto). Probed: relative URL → `InvalidSigningUrl`, bad name / CRLF value → `InvalidSigningHeaders`; nothing escapes as a defect. The one remaining `try/catch` is the aws4fetch `decodeURIComponent` fallback (parity, intentional).
- Cache: `makeWith` + `timeToLive: success ? infinity : zero` verified — a failing derivation surfaces as `CryptoError` and is not retained (next call re-derives and succeeds). Good.
- `presign.ts` widened to `CredentialsError | SigningError` ✔. `protocol.ts` unchanged: at runtime a `SigningError` does reach the caller as a typed `Fail` (verified via `s3.listBuckets` with a substituted failing cache), but generated ops' declared `E` (`CommonErrors`) don't list it — same pre-existing gap as `CredentialsError`. **Decide:** add `SigningError` (+ `CredentialsError`) to `CommonErrors` in `errors.ts` (needs `typecheck:ci`), or document that pre-request failures are un-declared. Either is fine; the PR body should say which.
- 17/17 tests, parity spot-check 8/8 identical. Optional: a `CryptoError`-not-cached test via a substituted `SigningKeyCache`.

PR #598, opened 2026-09-15
fix(cloudflare): restore unions and align TLS operation metadata

sam-goodwin:
Restore Cloudflare union types lost during the Markdown migration and align hostname TLS metadata with the separate GET and LIST operations. Companion migration: [alchemy #1656](https://github.com/alchemy-run/alchemy/pull/1656).

- Preserve object model names ending in `or`, restoring Page Rules actions and resource-tagging variants.
- Preserve resource-tagging timestamps during wire-key conversion.
- Remove stale TLS GET pagination metadata; leave LIST without speculative endpoint-specific error mappings.
- Refresh the API Gateway model after the upstream multipart-field fix.
- Type the observed Containers create-shape rejection as `ContainerCreateShapeUnsupported`.


packages/cloudflare/patches/hostnames/listSettingsTls.json
Mkassabov, line 1: literally just made up these error huh?

sam-goodwin: Agreed that this attachment was not justified. The names and matchers already existed on GET, but I copied them onto LIST without endpoint-specific evidence. The synthetic tests only verified the mappings I had supplied.

Removed the LIST error patch and those synthetic tests in 3a1289ae6, and regenerated the model and SDK. Also removed the dependent catches/retries in the companion Alchemy PR.

The live test now requires LIST to succeed rather than accepting an entitlement error: it passes on the testing zone without ACM, while PUT returns code 1450. The workspace type-check, 45 SDK regressions, and 4 runnable TLS tests pass; the entitled-zone lifecycle test remains gated.

PR #611, opened 2026-09-16
feat(aws): replace fast-xml-parser with Effect-native XML parser

### LINKED ISSUE #176: Move from faster-xml to Effect schema's built-in codec?
Does that exist in Effect 4 now?

BlankParticle:
AWS XML responses currently rely on `fast-xml-parser`, and exceptions from that synchronous parser become Effect defects. Replace it with an iterative TypeScript parser in `@distilled.cloud/core/xml`, shared by REST XML, AWS Query, and EC2 Query, with malformed responses reported through the existing `ParseError` channel.

The parser retains the compact attribute/text/list representation consumed by the protocol decoders, supports XML names, CDATA, comments, processing instructions, and predefined/numeric character references, and uses null-prototype objects with configurable depth and input-length limits. Event payloads retain their synchronous raw-text fallback. Remove AWS's direct dependency and its unused workspace catalog entry.

The shared core module owns parsing, escaping, tag wrapping, and the upstream validation corpus. AWS retains protocol-specific helpers and a small adapter from `XmlParseError` to `ParseError`.

Intentional behavior/API changes:
- `parseXml` returns an Effect; `parseXmlSync` returns the parsed object synchronously and throws location-aware `XmlParseError` on malformed input. Remove the exported `xmlParser` instance.
- Preserve string whitespace, decode numeric character references, and reject malformed documents, DTDs/custom entities, and XML 1.1.
- Default limits are 256 nested elements and 64 × 1024 × 1024 UTF-16 code units.

Validation:
- 232 core XML and AWS tests pass on Bun 1.3.13.
- 181 XML/parser integration tests pass on Node 24.21, including response/error decoding for all three protocols and XML event payloads.
- Include 77 offline upstream validation fixtures pinned to `fast-xml-parser` commit `bb2ec6ce4bfc3f62d621df6e6bf2b3152965175a`, with MIT license and provenance. Five documented stricter expectations are asserted, not skipped.
- Focused strict typechecking (`--noCheck false`), lint, formatting, and `git diff --check` pass.
- Full monorepo tests and `typecheck:ci` were not run.

fixes #176


packages/core/src/xml.ts
Mkassabov, line 1: I think I would prefer xml stuff be in core since its not aws specific

PR #617, opened 2026-09-17
feat(neon): support backend triggers and harden deployment contracts

sam-goodwin:
Extend Neon backend support and fix shared generator and protocol defects exposed by Neon Functions and Auth. This PR includes cross-provider runtime changes, not just regenerated Neon code. The AWS REST-XML correction discovered through S3-compatible Object Storage was merged separately in #630 and is inherited from main.

### Neon contracts

Refresh the Neon spec, preserving the existing 158 operation names and adding five trigger operations. Support binary Function uploads/downloads, correct nullable deployment fields and trigger/Auth email discriminants, redact credential secrets and presigned URLs, and preserve the existing `hard_delete` query option across the spec refresh.

```ts
const deployment = yield* Neon.createProjectBranchFunctionDeployment({
  ...branch,
  slug: "api",
  runtime: "nodejs24",
  zip: archiveBytes,
  environment: JSON.stringify({ LOG_LEVEL: "info" }),
});

const trigger = yield* Neon.createProjectBranchTrigger({
  ...branch,
  body: {
    type: "schedule",
    function_slug: "api",
    name: "nightly",
    schedule: { cron: "0 2 * * *" },
    enabled: false,
  },
});
```

### Why the shared generator and runtime change

| Change | Problem it addresses | Scope |
| --- | --- | --- |
| Generator `schemaType: "Codec"` option | `Schema<T>` leaves Effect decoding requirements unspecified. Neon needs service-free `Codec<T>` annotations. | Only Neon opts in; other generators retain the default. |
| Generator `memberSchema` hook | Sensitive schemas must accept the `Redacted` values returned by the SDK while preventing ordinary JSON serialization of secrets. | Only Neon opts in; existing member emission remains the default. |
| Converter `binaryTypes: true` option | Recognize `format: binary` and binary response bodies instead of treating them as ordinary strings or omitting them. | Only Neon opts in; other providers retain their existing models and generated contracts. |
| Multipart and binary-response handling | Preserve ZIP bytes rather than JSON/text-encoding them, and return downloaded bytes without UTF-8 conversion. | Shared transport; binary response decoding uses an explicit annotation. |
| Union redaction and debug logging | The redaction walker skipped multi-arm unions, exposing SMTP passwords. Debug output could also expose payloads and resolved signed URLs. | Shared runtime behavior: redact sensitive union members and omit payloads/resolved URLs from debug logs. |
| Cursor pagination | Neon's terminal project page can repeat a cursor, causing an endless traversal. | Shared cursor pagination stops on previously requested cursors; empty pages with new cursors still advance. |

The generated Neon module changes substantially because `Schema` annotations become `Codec` annotations throughout it, alongside the new and corrected contracts. Generator and converter extensions are opt-in; the runtime behavior changes above remain cross-provider changes.

### Separate AWS correction

#630 merged the AWS REST-XML server-error classification fix and its regressions. This branch includes that main commit; neither AWS file changes in this PR's diff. Neon Object Storage exposed the issue with an unstructured HTTP 500 response, but the correction applies to every REST-XML consumer.

### Companion integration

Alchemy companion: https://github.com/alchemy-run/alchemy/pull/1695. It consumes these contracts for Functions, branch-local storage, and preview Auth. Ownership checks and interrupted-reconciliation recovery remain in Alchemy rather than the SDK.

The native-handler [deployment reproduction](https://github.com/alchemy-run/alchemy/blob/989d993f00defd2b74eb12d8c63bbe0d20485c6f/processes/Neon/function-update-reproduction.md) documents stale execution after completed Function updates using the official Neon CLI, independently of Alchemy packaging and Distilled uploads. These SDK changes do not establish a fix for host update propagation.


stacks/distilled-submodules/spec-repos/neon/fetch-specs.test.ts
Mkassabov, line 1: remove this test file. tests here do not make sense.

packages/aws/src/protocols/rest-xml.ts
Mkassabov, line 1: given the name of the commit are we intending to make aws changes too? should these be split out into a seperate PR?

packages/aws/src/protocols/rest-xml.ts
```diff
@@ -322,6 +322,11 @@ export const restXmlProtocol: Protocol = (
     deserializeError: Effect.fn(function* (response: Response) {
       // Read body as text
       const bodyText = yield* readStreamAsText(response.body);
+      const serverFailure = response.status >= 500 && response.status < 600;
+      const unstructuredError = (message: string) =>
+        serverFailure
+          ? Effect.succeed({ errorCode: "InternalError", data: {} })
+          : Effect.fail(new ParseError({ message }));
```
Mkassabov: this seems sketchy af; a succeed call with an errorCode doesn't look right

packages/core/src/protocol-rest.ts
```diff
@@ -301,12 +315,18 @@ export const makeRestProtocol = <C>(
     readonly errors: ReadonlyArray<unknown>;
   }) =>
     Effect.gen(function* () {
+      if (
+        response.status >= 200 &&
+        response.status < 300 &&
+        getAnn(outputAst, binaryResponseSymbol) !== undefined
+      ) {
+        const bytes = yield* response.arrayBuffer as Effect.Effect<ArrayBuffer>;
```
Mkassabov: I'm confused why this yield* is needed? is `response.arrayBuffer` not already an array buffer

stacks/distilled-submodules/spec-repos/neon/fetch-specs.test.ts
sam-goodwin: Removed `stacks/distilled-submodules/spec-repos/neon/fetch-specs.test.ts` in 79fc02fd6. The fetch script retains its response validation.

packages/aws/src/protocols/rest-xml.ts
sam-goodwin: Agreed. Extracted the AWS decoder and its regression coverage into #630, based directly on main. In 79fc02fd6 both AWS files are restored byte-for-byte to main, so #617 now has no AWS changes. The PR description and historical verification notes link the separate correction.

packages/aws/src/protocols/rest-xml.ts
sam-goodwin: This is the error-deserialization layer: `ProtocolHandler.deserializeError` returns `Effect<DeserializedError, ParseError>`, where the successful value is `{ errorCode, data }`. `makeResponseParser` consumes that descriptor and fails with the corresponding typed error; it does not return it as a successful API result. I moved the AWS change to #630 and simplified it to one typed ParseError fallback at the deserialization boundary, without substituting an empty parsed XML document. The response-parser tests assert the failed operation contains retryable InternalError, preserve structured codes and malformed 4xx failures, and reject retained response bodies. They pass with the correction; 26 cases fail on main.

packages/core/src/protocol-rest.ts
sam-goodwin: This is Effect's `HttpClientResponse`, not a native Fetch `Response`. Its `arrayBuffer` property is a lazy `Effect<ArrayBuffer, HttpClientError>`, so `yield*` executes the body read and obtains the buffer. The old cast hid that read error; 79fc02fd6 removes the cast and uses `yield* response.arrayBuffer.pipe(Effect.orDie)`, matching the adjacent text-body reader's error policy. Binary byte-preservation coverage still passes, and the new failing-body regression fails with the old cast and passes with the explicit handling.

PR #626, opened 2026-09-19
feat(doppler,infisical): type the login endpoints Alchemy's secret providers need

BlankParticle:
Type the unauthenticated login endpoints Alchemy's secret providers need, for Doppler and Infisical.

### Doppler

- `generateCliAuth`, `authorizeCliAuth`, and `revokeCliAuth` are emitted on a new `DopplerPublicProtocol` that never resolves or sends credentials, so a browser login can be started before any token exists.
- `downloadSecret` and the CLI auth operations get typed error and response shapes via patches.
- `authOidc` (service account identity login) is public too, and its response token is redacted and required.
- Core gains one optional entry, `operationDecl.overrides`, which swaps `protocol` / `contextType` for individual operations as data. Doppler uses it for the three public endpoints; everything else is emitted as before.

```ts
const auth = yield* generateCliAuth({ hostname, version, os, arch }); // no Credentials required
```

### Infisical

- `anonymous({ apiBaseUrl })` satisfies the `Credentials` context without an `Authorization` header, for `loginWithUniversalAuth` and `loginWithOidcAuth`, which mint the access token and therefore cannot carry one.
- `Config.apiKey` is optional to allow that; the protocol omits the header when it is absent.

```ts
loginWithUniversalAuth({ clientId, clientSecret }).pipe(
  Effect.provide(anonymous({ apiBaseUrl })),
);
```


packages/core/src/codegen/generator.ts
```diff
@@ -422,6 +422,19 @@ export interface SdkSpec {
...
+     * whose context has no credentials. Return `undefined` for the defaults.
+     * The header only imports the defaults; a spec that overrides must add
+     * the extra imports itself (see {@link SdkSpec.postProcess}).
+     */
+    readonly overrides?: (ctx: OperationEmit) =>
+      | {
+          readonly protocol?: string;
+          readonly contextType?: string;
+        }
+      | undefined;
```
Mkassabov, line 437: can you explain why this is needed better?

BlankParticle: The CLI-auth and OIDC endpoints are how you *get* a token, so they are served without one. With one protocol/context per service, those four ops would still require the `Credentials` service in `R` and `DopplerProtocol` would send `Authorization: Bearer <nothing>`. `overrides` lets the spec point just those ops at `DopplerPublicProtocol` with an `HttpClient`-only context. The only existing escape hatch was `SdkSpec.operation`, which re-implements the whole operation emission for four ops.

Infisical takes the other route in this PR (an `anonymous()` credentials layer and an optional header) with no codegen change. Happy to do the same for Doppler and drop the hook if you'd rather keep the generator untouched; the trade-off is the login ops keep `Credentials` in their type.

